PR #16, opened 2026-02-13
add stop feature from RC's joystick or UI fix #15

### LINKED ISSUE #15: Implement a safety feature to stop all processess and return the control to the pilot
This is an essential feature to prevent the aircraft to fly or perform any commanded action and lock the pilot's RC commands. Appears that during a flight mission or any automated control command, the control signal from the RC joysticks is disable until finish the specified action.
Hence, an emergency stop process must be triggered if:

1. An RC joystick channel or more are a certain threshold value away the center. This options seems better to be a user-setting if wants that behavior.
2. The stop button from the `MainActivity` was pressed.

angel-ayala:
The following changes aims to solve #15 and the architecture was affedted. `AircraftHandler` now act as entry point for all SDK drone-related interaction. `TelemetryHandler` and `MissionHandler` now are contained by `AircraftHandler` for easy mission and commands management.

A full start/stop and pause/resume test is needed to ensure a good functioning, maybe this can me merged and create new issues for possible bugs.

@mjohenneken you can accept the merge if you don't have any comments

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/TelemetryData.kt
+++ b/app/src/main/java/org/WenuLink/adapters/TelemetryData.kt
@@ -40,7 +40,26 @@ data class RCData(
     val buttonC2: Boolean,
     val buttonC3: Boolean,
     val mode: FlightModeSwitch?
-)
+) {
+    fun toMAVLink(): RCData {
+        val currRC = this.copy(
+            throttleSetting = (this.throttleSetting + 660) / 1320,  // TODO: check this value
+            // Mavlink: 1000 to 2000 with 1500 = 1.5ms as center...
+            leftStickVertical = (this.leftStickVertical * 0.8).toInt() + 1500,
+            leftStickHorizontal = (this.leftStickHorizontal * 0.8).toInt() + 1500,
+            rightStickVertical = (this.rightStickVertical * 0.8).toInt() + 1500,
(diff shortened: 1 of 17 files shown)


app/src/main/java/org/WenuLink/adapters/TelemetryData.kt
```diff
@@ -40,7 +40,26 @@ data class RCData(
     val buttonC2: Boolean,
     val buttonC3: Boolean,
     val mode: FlightModeSwitch?
-)
+) {
+    fun toMAVLink(): RCData {
+        val currRC = this.copy(
+            throttleSetting = (this.throttleSetting + 660) / 1320,  // TODO: check this value
```
mjohenneken: Not clear What are these numbers? Suggestion: Comment, Named Constants or separate conversion function

mjohenneken: ahh saw other comment // DJI: range [-660,660]

mjohenneken (dismissed):
Looks beautiful! 

The Virtual Stick conversion is the only thing i got stuck during reading the code.

In general the code base looks super clean with the architecture we've been following.

mjohenneken:
It seems I need write access for the automatic checker to continue

angel-ayala:
I just added some comments and fixed the transformation function, I just checked and now is ok. Also I update your role with write access.

mjohenneken (approved):

Issue #13, opened 2026-02-05
Implement a boot sequence for proper initialization


angel-ayala:
Current skeleton only defines an example of boot sequence that needs to be improved, specially to have a proper ready-to-fly status report, in some tests, when moving too fast from the service execution to takeoff, logcat's logging correctly reported the error due to sensor state but with no action on the application.

Additionally, must consider a hot startup, in the case that the application unexpectedly exits, can start over even if the aircraft already take off.

Initial version can address the following SDK components, [DJICompass](https://developer.dji.com/api-reference/android-api/Components/Compass/DJICompass.html), [DJIIMUState](https://developer.dji.com/api-reference/android-api/Components/IMUState/DJIIMUState.html), and [DJIGravityCenterState](https://developer.dji.com/api-reference/android-api/Components/FlightController/DJIFlightController_DJIGravityCenterState.html).

If any other component appears to be essential, please report.

Issue #14, opened 2026-02-05
Implement the Camera Protocol


angel-ayala:
Implement the required controller to deal with the microservice of the [Camera Protocol](https://mavlink.io/en/services/camera.html). Main focus is to enable the photogrammetry use case.

PR #17, opened 2026-02-17
Enhance boot sequence for proper init/shutdown process fix

angel-ayala:
This PR address #13 to enhance the Coroutine lifecycle's widely reducing the asynchronous calls replaced by suspend and waiting methods. Additionally, sensor calibration state and proper in-flight initialization of `AircraftStateMachine`.

Now the boot sequence starts with `WenuLinkService`  and the `AircraftHandler` initialization and boot.
`MAVLinkService` and `WebRTCService` comprise their own boot/unload sequence. When termination a timeout prevent for infinite coroutine execution by cancelling the scope.


TimonSchreiber:
While reviewing FCManager I noticed `toString()` has the connected/disconnected branches swapped.

TimonSchreiber:
# Test Report: Hardware & Software

- Drone 1: DJI Mavic Pro with DJI GL200A
- Drone 2: DJI Mavic 2 Pro with DJI RC1B
- Android Device: Lenovo TB-X606X (Android 10)
- GCS: QGroundControl v5.0.8 (64-bit)
- Note: Testing conducted indoors without rotors. Wireless ADB used for simultaneous Logcat access.

## Findings
Three bugs were identified that prevented successful connection. All three required local patches before any meaningful MAVLink communication was possible. The patches are described below for reference.

### Bug 1: IMU sensor state read too early in `checkSensors()` (Mavic 2 Pro)
`checkSensors()` calls `getIMUState()` immediately after `isReadingSensors()` returns true, approximately 1 second after the telemetry listener is registered. At that point the IMU sensors are still in a `BOOT` state rather than `OK`, causing both accelerometer and gyroscope checks to fail and boot to abort.
Logcat evidence:

> 14:05:56.025  AircraftHandler  Init. telemetry: OK
> 14:05:56.026  AircraftHandler  Waiting sensors data
> 14:05:57.048  AircraftHandler  Accelerometer error!
> 14:05:57.048  AircraftHandler  Gyroscope error!

The compass (checked via a separate path) showed `hasError: false, calState: NOT_CALIBRATING` at the same timestamp, confirming the sensors were physically functional, only the IMU values had not yet settled.
Local fix applied: replaced the immediate `getIMUState()` call with a `waitTimeout` poll (200ms interval, 10s timeout) that waits for all IMU states to reach `OK` before evaluating.

### Bug 2: `waitHomeUpdate()` blocks indefinitely without GPS
After a successful boot, `waitHomeUpdate()` calls `updateHomeCoordinatesFromAircraft()` in a loop via `AsyncUtils.waitReady()` with no timeout argument. Indoors with weak GPS, `FCManager.setHomePosition()` fails on every iteration, and the loop never exits, permanently blocking `boot()` from returning and preventing `runServices()` from ever being called.
Logcat evidence (repeating at 1Hz until manually stopped):

> 14:22:35.263  AircraftHandler  Aircraft boot: OK
> 14:22:35.263  AircraftHandler  Requesting home coordinates update...
> 14:22:35.267  AircraftHandler  Error request: GPS level is too weak...
> 14:22:36.267  AircraftHandler  Requesting home coordinates update...
> 14:22:36.268  AircraftHandler  Error request: GPS level is too weak...
> [repeats indefinitely]

Local fix applied: replaced `waitReady()` with `waitTimeout(1000L, 30000L, ::updateHomeCoordinatesFromAircraft)` and added a warning log on timeout. Boot continues regardless of whether home was set, consistent with how all other boot steps handle failure.

### Bug 3: `isSystemReady()` requires home position, blocking telemetry unlock indoors
With bugs 1 and 2 fixed, `boot()` completed and MAVLink started successfully. GCS detection worked, parameters were exchanged, and `MISSION_ACK` was received. However `waitSystemReady()` never resolved because `isSystemReady()` requires `aircraft.state.isHomeSet()` in addition to `navigation.wasRequested` and `parameter.wasRequested`. Without GPS, `isHomeSet()` is always false, causing `waitSystemReady()` to time out after 60 seconds and `notifySystemReady()` to never be called, permanently keeping `readOnlyMessageRate = true` and suppressing all non-heartbeat telemetry.
Logcat evidence:

> 14:43:54.900  MAVLinkController  GCS found?: true
> 14:43:54.900  MAVLinkController  Waiting for parameters
> 14:43:55.929  MAVLinkService     Parameters loaded
> 14:43:55.931  MAVLinkController  Waiting for MAVLink microservices.
> 14:44:07.729  MAVLinkController  Processing message: MAVLINK_MSG_ID_PARAM_REQUEST_LIST
> 14:44:07.865  MAVLinkController  Processing message: MAVLINK_MSG_ID_MISSION_REQUEST_LIST
> 14:44:07.885  MAVLinkController  Processing message: MAVLINK_MSG_ID_MISSION_ACK
> [no notifySystemReady — waitSystemReady times out 60s after microservices line]

A home position is a flight safety prerequisite for arming, not a MAVLink protocol readiness condition. Telemetry should flow regardless of GPS availability.
Local fix applied: removed `aircraft.state.isHomeSet()` from `isSystemReady()`. After this change, `ready?=true` was confirmed in the log for the first time:

> 15:17:03.415  MAVLinkController  MAVLink microservices initialized.
> 15:17:03.415  MAVLinkService     MAVLinkService (ready?=true) (GCS?=true) (listening?=true) (sending=true)

### Remaining Issue: No telemetry displayed in QGC after all three fixes
With all three local patches applied, the service reached `ready?=true` and remained stable. QGC was actively negotiating: continuous `REQUEST_DATA_STREAM` messages were received and `processMessageInterval (COMMAND_LONG ID 511)` was being handled correctly, updating the `messageRates` list with QGC's requested stream IDs and intervals. However, no telemetry was displayed in QGC.

angel-ayala:
Really nice report @TimonSchreiber , I'm aware of all the situations and now that you mention it, the major problem  here are current waiting mechanisms. Since most test I did used the Simulation feature of the SDK, never took into account the GPS issue on indoor environments to mission upload/download or any other pre-flight operations.

I will update the code considering such cases as you mention.

Regards to the Telemetry displaying, the application is not reporting anything since `messageRates` is in read-only mode until `isSystemReady()`. So Whenever a `REQUEST_DATA_STREAM` is received, it will only update the `messageRates` afterwards. The expected behavior was to only send `HEARTBEAT` messages until the parameters and mission items were requested, as [communication flow](https://docs.qgroundcontrol.com/Stable_V5.0/en/qgc-dev-guide/communication_flow.html) explains. Additionally, when debugging the network traffic I discovered that no telemetry message is sent before asking for parameters or telemetry. That logic seems OK to me, but definitively I needs to improve the implementation.

I will try to implement the changes today or tomorrow, I'll let you know.

angel-ayala:
Major refactors were made, explained next:
- Each aircraft's command will take a variable amount of time and must run under the main scope, without blocking the mavlink process in any sense. hence, the Command, Single Consumer Queue, State Machine, and Guard patterns were adopted to address a maintable code.
- Now, when `WenuLinkService` starts, it will instanciate `AircraftHandler` and trigger the `BootCommand`, starting a simplified boot sequence of `processUserPreferences`, `startTelemetry`, to update state and sensors readings in manual control mode.
- Similarly, from any `AircraftCommand` such as `ArmCommand`, `DisarmCommand`, `TakeoffCommand`, `LandCommand` in `adapters.aircraft.AircraiftCommands.kt`, with properly defined `validate(...)` and `execute(...)` methods will be executed. For `AircraftState`, same patterns were applied with `canTransition(...)` and `reduce(...)`.
- Since all sensors readings are updated over time, and therefore their state, a background coroutine is running for async polling the current state and guarding specific cases. For example, the safety logic checking if the user move any joystick by listening to `AircraftHandler.rcInput` to make the aircraft force manual control if flying. The implemented `guardJob` loop handles such situation in a reactive manner, encapulating state conditioned behavior inside `AircraftHanlder`. 
- Nevertheless, additional refactors must be addressed for proper flight mode and aircraft state transitions, mainly for file organization.

**TESTS ARE NEEDED**

**General observations**

Current changes should address the endless wait for sensor data and home position, it must alleviate the safety-layer by centralizing the calls in `AircraftHandler` for direct status reading and command requests. A modularized version for async commands must also ensure readibility and maintainability, `AircraftCommand` ended to be similar to `MissionAction` such as the case of missions. Moreover, during diverse tests of the mission service, unexpected behaviors suddenly appeared, solved mostly by starting the application all over again. Hence, the issue could possibly to be related with async timing. The mission service must be updated and reviewed to check if possible classes can be reused or refactored for corretly address Waypoints and Event missions. Current refactor should also alleviates the inclusion of any other missing MAVLink's microservice such as the case of the Camera service requested in #14, due the Command pattern.

Another major final refactor I see necessary, is related to file organization, requiring to migrate `TelemetryHandler`, `TelemetryData`, `AircraftHandler`, and `AircraftState` to the `aircraft` package. In addition to `MissionHandler` be migrated to `mission`.

app/src/main/java/org/WenuLink/adapters/AircraftHandler.kt
```diff
@@ -49,118 +50,203 @@ class AircraftHandler {
     var isPowerOff = false
```
TimonSchreiber:
```suggestion
    var isPowerOff = true
```

app/src/main/java/org/WenuLink/adapters/AircraftHandler.kt
```diff
@@ -179,76 +265,72 @@ class AircraftHandler {
...
+                    // --- Enforce consistency ---
+                    enforceModeConsistency()
 
-        logger.i { "Aircraft in standby" }
-        // All seems ok
-        stateTransition(AircraftState.Standby)
+                    // --- Safety hooks ---
+                    rcInput?.hasCenteredJoystick()?.let {
+                        // stop everything to if any joystick is moved
+                        manualControl()
```
TimonSchreiber:
```suggestion
                        if (!it) manualControl()
```

TimonSchreiber (changes requested):
Hi, I reviewed the patch and tested it with the Mavic 2 Pro, indoors without rotors as before. The architectural direction is genuinely good, the Command pattern, the state machine with `canTransition`/`reduce`, and centralising the guard loop in `AircraftHandler` are all great improvements over the previous design. The changes to `waitServicesRequest` also correctly fix the telemetry rate-lock issue from my earlier report. Below are the issues I found, split by severity.

### Bugs confirmed by logcat
#### 1. `isPowerOff` initial value `false` should be `true` (see inline suggestion)
`BootCommand.validate` rejects the boot with "Already booted" when `isPowerOff` is `false`, which is its initial value. Since the aircraft starts in an unbooted state, the field should be `true` at the beginning. The same initial value also breaks `isAircraftReady()` in `WenuLinkService`, which would return `true` immediately before boot completes. I tested with `var isPowerIff = true` and the MAVLink service stared correctly ~650 ms after booting.

#### 2. Joystick safety hook ignores the boolean, guard triggers `manualControl()` unconditionally (See inline suggestion)
```kotlin
// current: calls manualControl() whenever rcInput is not null
rcInput?.hasCenteredJoystick()?.let {
    manualControl()
}

// should be
rcInput?.hasCenteredJoystick()?.let {
    if (!it) manualControl()
}
```

#### 3. Double state dispatch in boot, takeOff, and land
`BootCommand.execute` dispatches `BootAircraftState`, then calls `boot()`, which calls `applyState(BootAircraftState)` again. Same pattern in `TakeoffCommand`/`takeOff()` and `LandCommand`/`land()`. The second dispatch fails `canTransition` silently, confirmed in logcat:
```
Executing: BootCommand
Invalid state transition: ...BootAircraftState@e5cf7d2
Aircraft booting...
```
Since the Commands now own the state transition, the `applyState(...)` calls inside `boot()`, `takeOff()`, and `land()` should be removed.

### Possible bugs
#### 4. `GoHomeCommand` dispatches `FlightTerminationAircraftState`
RTL is a normal, recoverable operation. `FlightTerminationAircraftState` maps to `MAV_STATE_FLIGHT_TERMINATION`, which signals an emergency abort to QGC. This may be unintentional, if so a dedicated `ReturningHomeAircraftState` or reuse of `FlyingAircraftState` might be more appropriate.

#### 5. `sensorsHealty` may be `false` during the first `ArmCommand`
`sensorsHealthy` is only updated by the guard loop, which starts after `registerHandlerScope` is called. There is a window after boot where `sensorsHealthy` is `false` not because sensors are bad, but because the guard loop hasn't had its first iteration yet. One option is to run `sensorsCheck()` once at the end of `boot()` to ensure the field has a real value before the aircraft becomes ready.

#### 6. No standby transition after landing
`LandCommand` waits for `awaitFlightState(false)` but does not follow up with a `StandbyAircraftState` dispatch or disarm. The aircraft would remain in `MAV_STATE_ACTIVE` on the ground indefinitely. Auto-dispatching `DisarmCommand` after confirmed landing would close this gap.

### Design notes
#### 7. Guard loop sensor polling is too aggressive
`sensorChecks(1000L)` runs a full blocking wait cycle, including `waitHomeSet`, on every guard iteration. With GPS unavailable indoors, this generated roughly 8–10 failed `setHomePosition` SDK calls per second continuously, visible in logcat. The guard loop is a safety check cycle, not a sensor initialisation loop. Reading from already-cached values (as `isCompassOk()` does via `FCManager.compassOk()`) or separating the sensor health refresh into a slower background job would keep the guard cycle predictable and avoid polluting the SDK.

#### 8. `AircraftCommand.validate` checks state machine values
`validate` in `ArmCommand` checks `ctx.state.isStandBy()` and `ctx.state.isOnTheGround()`, but `canTransition` in `ArmAircraftState` already enforces exactly those conditions. The two layers are meant to guard different things:
| Method | Question | Examples |
| --- | --- | --- |
| `validate` | Can the *hardware* do this? | Sensors health, GPS, SDK connected |
| `canTransition` | Is this *logically valid* from current state? | Standby before arm, fyling before land |

Mixing state machine checks into `validate` means the same condition has to be kept in sync in two places.

#### 9. `currentCommand` field is set but never read
It's assigned in the command processor and cleared in `finally`, but nothing reads it, not the guard loop, UI, or any validation path. If it's intended for status display or cancellation, it appears incomplete. If not needed, it adds noise.

#### 10. Naming `AircraftState` / `AircraftStateMachine.UnifiedState`
The sealed interface `AircraftState` represents the intent to transition, not the current state, which makes the name confusing, especially since `UnifiedState` is the thing that actually holds current state. Suggested rename: `AircraftState` → `AircraftTransition`, with concrete names like `BootTransition`, `ArmTransition`, `TakeoffTransition`, `LandTransition`...

`UnifiedState` could then be renamed to `AircraftState`, so the dispatch signature reads naturally: `dispatch(event: AircraftTransition): Result<AircraftState>`.

---

### Out of scope
I have additional observations around code style, readability, and a couple minor issues in unrelated classes like `FCManager`. I have set those aside for now, will bring them up later if they still apply after this is merged.

angel-ayala:
Definitevly N1, N2, and N3 where bugs and already fixed.

Regards to Possible bugs:
- N4 I changed the state to `FlightTransition`.
- N5 I do not see it as a problem, but make sense to include an initial check on bot.
- N6 Really good catch! I got it, but when passing to the new architecture was left behind, now the `LandCommand` triggers `DisarmCommand` after confirmation in touching ground.

Regards to Design note:

- N7 I totally agree with the comments, differents rates for checking sensor and other information was implemented to control this frequency. The fastest is `safetyChecks()` followed with `syncState()`, `sensorChecks()`, and the longest waiting for `updateHomePosition()`.
- N8 I did noticed the double validation, changes were made accordingly with most of command addressing returnin null for `validate()`.
- N9 `currentCommand` now is used to verify if can send a command or not.
- N10 A total refactor was don for `AircraftState` now being a dataclass with useful auxiliary functions. Added a clear separation between `state` and `stateMachine` with the first a read-only of vehicle's information and the second a interfaz to change such information.


PS: I used N in replace of Number, since the # character is used to reference issues.

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
```diff
@@ -0,0 +1,541 @@
...
+            }
+
+            setMode(fallbackMode)
+        } else {
+            syncBaseMode()
+        }
+    }
+
+    suspend fun syncState(sensorsInterval: Long = 1000L, homeInterval: Long = 5000L) {
+        if (!isPowerOff) return
```
```suggestion
        if (isPowerOff) return
```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
```diff
@@ -0,0 +1,541 @@
...
+        // Stop telemetry process
+        telemetry.launchTelemetry(false)
+        // Second wait to remove the existing data
+        return telemetry.waitDataRemoving(delay)
+    }
+
+    suspend fun waitBoot(timeout: Long = 5000L): Boolean = AsyncUtils.waitTimeout(
+        500,
+        timeout
+    ) { isPowerOff }
```
```suggestion
    ) { !isPowerOff }
```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
```diff
@@ -0,0 +1,151 @@
...
+            .onFailure { e ->
+                onResult(e.message)
+            }
+    }
+}
+
+data class RepositionCommand(val targetCoordinates: Coordinates3D, val speed: Float?) :
+    AircraftCommand {
+
+//    override suspend fun validate(ctx: AircraftHandler): String? =
```
TimonSchreiber:
```suggestion

```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
```diff
@@ -0,0 +1,151 @@
...
+//        if (!ctx.state.isFlying()) "Not flying" else null
```
TimonSchreiber:
```suggestion

```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
```diff
@@ -0,0 +1,151 @@
...
+            }
+            .onFailure { e ->
+                onResult(e.message)
+            }
+    }
+}
+
+object GoHomeCommand : AircraftCommand {
+
+//    override suspend fun validate(ctx: AircraftHandler): String? =
```
TimonSchreiber:
```suggestion

```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
```diff
@@ -0,0 +1,151 @@
...
+//        if (!ctx.state.isFlying()) "Not flying" else null
```
TimonSchreiber:
```suggestion

```

TimonSchreiber (commented):
Great work as always @angel-ayala !

### Bugs:

`waitBoot` and `syncState` have inverted `isPowerOff` guards (also noted as inline suggestions). In `waitBoot`, the lambda `{ isPowerOff }` is passed as the `isReady` condition to `AsyncUtils.waitTimeout`, which loops while `!isReady()`. This means the function exits when `isPowerOff` is `true`, exactly when the aircraft is off, so it returns immediately on startup and never actually waits for boot to complete. It should be `{ !isPowerOff }`.
In `syncState`, the guard `if (!isPowerOff) return` skips the entire sync when the aircraft is running, which is the opposite of what is intended. It should be `if (isPowerOff) return`. These two bugs together mean `WenuLinkService` starts MAVLink before boot completes and the monitor loop skips state synchronisation entirely during flight.

`processLanding` disarm will always be silently dropped. The current code dispatches `DisarmCommand` from inside the `LandCommand` completion callback:
```kotlin
aircraft.dispatchCommand(LandCommand()) { landError ->
    if (landError == null) {
        aircraft.dispatchCommand(DisarmCommand()) { ... }
    }
}
```

The problem is that `onResult` is called from inside `startCommandProcessor`'s `for` loop, in the `execute` coroutine, **before** the `finally { currentCommand = null }` block runs. When `dispatchCommand(DisarmCommand())` is called, `currentCommand` is still set to `LandCommand`, so the busy guard fires and the disarm is silently skipped every time.
A better way might be to have `LandCommand.execute` handle the disarm itself rather than delegating it upward through the callback. After `awaitFlightState(false)` confirms the aircraft is on the ground, disarm directly:
```kotlin
data class LandCommand(val forceConfirmation: Boolean = true) : AircraftCommand {
    override suspend fun execute(ctx: AircraftHandler, onResult: (String?) -> Unit) {
        ctx.stateMachine.dispatch(LandTransition)
            .onSuccess {
                ctx.land()
                val onTheGround = ctx.awaitFlightState(false)
                if (!onTheGround) {
                    onResult("Unable to land")
                    return
                }
                ctx.disarmMotors()  // <- new
                ctx.waitArmTransition(false, 5000L) // <- new
                onResult(null)
            }
            .onFailure { e -> onResult(e.message) }
    }
}
```
`CommandController.processLanding` then simply dispatches `LandCommand()` with a single callback for logging, and the chained `DisarmCommand` dispatch can be removed.

`RepositionCommand` and `GoHomeCommand` have dead commented-out `validate` methods. These should be removed entirely rather than left commented out. If validation needs to be added later, a `TODO: ` comment with a brief explanation is much clearer.

## Suggestion:
`sensorChecks` divides the timeout by 3, but the number of timed checks is not enforced by the code. Currently there are exactly 3 `waitTimeout` calls using `perSensorTime`, so the arithmetic is correct today, but it is fragile. Adding or removing a sensor check without updating the divisor would silently miscalculate every timeout. Tying the divisor to the actual number of checks would make this self-correcting.

A suggestion:
```kotlin
suspend fun sensorChecks(timeout: Long = 10000L): Boolean {
    logger.i { "Waiting sensors data" }
    if (!AsyncUtils.waitTimeout(100L, timeout, isReady = telemetry::isReadingSensors)) {
        logger.e { "No sensor readings!" }
        return false
    }
    val checks: List<suspend (Long) -> Boolean> = listOf(
        { t -> waitCompass(t) },
        { t -> waitAccelerometer(t) },
        { t -> waitGyroscope(t) }
    )
    val perSensorTime = (timeout.toFloat() / checks.size).roundToLong()
    var allOk = true
    for (check in checks) {
        if (!check(perSensorTime)) allOk = false
    }
    return allOk
}
```

Everthing else looks good to me!

angel-ayala:
Thanks @TimonSchreiber , I just addressed the bugs and other architectural changes.
Giving a second thought, `AircraftCommand` should be validate if can transit to its corresponding `StateTransition`, makes no sense to report a failure on the command execution when the problem was an incompatible state.
Therefore, now the commands validates the state transition and aircraft capacities.

Additionally, regards the command enqueue, the `currentCommand` attribute aimed to check if another command can be executed but does not aid  to stop current commands. Instead, a `currentCommandJob` is now used to wrap the `AircraftCommand.execute(...)`  logic inside a job, that can be canceled from `AircraftHandler.stopCommand()`.
Now, `AircraftHandler.dispatchCommand(...)` can be called successfully enqueue commands for the transition case of  `LandingCommand` -> `DisarmCommand`.
The use of `AircraftCommand`s should be prioritized since since it keeps `AircraftStateMachine` updated.

Hopefully, now all should be properly addressed.

I must be sincere here, now revisiting the code, there was a lot of logic implemented by a naive instinct since all the coroutine aspects and async calls was new for me developing in Kotlin. So possibly, there are other methods that can be more simple than can em enhanced in future versions.

TimonSchreiber (commented):
### Unresolved:

The two `isPowerOff` bugs from the previous review are still present. `waitBoot` in `AircraftHandler` still passes `{ isPowerOff }` as the readiness predicate, which exits immediately when the aircraft is off rather than waiting for boot to complete. `syncState` still has `if (!isPowerOff) return`, which skips state synchronisation during normal operation. Both need to be inverted!

### Bug: `InitialTransition.canTransition` is always non-null:

```kotlin
override fun canTransition(from: AircraftState): String? = when {
    from.mavlink != MAV_STATE.MAV_STATE_BOOT -> "Not from boot"
    from.mavlink != MAV_STATE.MAV_STATE_POWEROFF -> "Not from power off"
    else -> null
}
```
The second branch is unreachable in any meaningful sense. For it to be evaluated, the first branch must have been false, meaning `from.mavlink == MAV_STATE_BOOT`. At that point `from.mavlink != MAV_STATE_POWEROFF` is always true, so the function always returns a non-null string and `canDispatch(InitialTransition)` always rejects. The `else -> null` branch is dead code.
The fix is to use a value-based `when`, which is also the more idiomatic Kotlin form:
```kotlin
override fun canTransition(from: AircraftState): String? = when (from.mavlink) {
    MAV_STATE.MAV_STATE_BOOT,
    MAV_STATE.MAV_STATE_POWEROFF -> null
    else -> "Cannot reset to initial from current state"
}
```
Note that because `dispatch` now applies `reduce` unconditionally, `BootCommand.onStop` calling `stateMachine.dispatch(InitialTransition)` directly will still execute the reset regardless of this bug, so the reset itself works at runtime. But any `validate()` path going through `canDispatch(InitialTransition)` would always be blocked, which is incorrect and confusing.

#### Suggestion: missing state helper methods create inconsistent style

Most transitions use readable helpers like `from.isFlying()`, `from.isArmed()`, and `from.isOnTheGround()`. `BootTransition` and `InitialTransition` are forced to compare `from.mavlink` directly because `AircraftState` has no helpers for `UNINIT`, `BOOT`, or `POWEROFF`. Adding the three missing helpers would let every transition use the same style:
```kotlin
fun isUninitialized() = mavlink == MAV_STATE.MAV_STATE_UNINIT
fun isBoot() = mavlink == MAV_STATE.MAV_STATE_BOOT
fun isPowerOff() = mavlink == MAV_STATE.MAV_STATE_POWEROFF
```
Then `InitialTransition` and `BootTransition` become consistent with the rest:
```kotlin
override fun canTransition(from: AircraftState): String? = when {
    from.isBoot() || from.isPowerOff() -> null
    else -> "Cannot reset to initial from current state"
}

override fun canTransition(from: AircraftState): String? = when {
    !from.isUninitialized()) -> "Already initialized"
    else -> null
}
```

TimonSchreiber (commented):
Everything else looks good

angel-ayala:
Sorry about that @TimonSchreiber , I had the `isPowerOff` related changes done but then I find out that I was on another branch so I stashed the changes , but I forgot to pop such changes and applied it. Now it be OK. I also addressed the other bugs and suggestions you reported.

I didn't mentioned before but your earlier suggestion:
>`sensorChecks` divides the timeout by 3, but the number of timed checks is not enforced by the code. Currently there are exactly 3 `waitTimeout` calls using `perSensorTime`, so the arithmetic is correct today, but it is fragile. Adding or removing a sensor check without updating the divisor would silently miscalculate every timeout. Tying the divisor to the actual number of checks would make this self-correcting.

It was not addressed since such adaptive behavior fits best in those cases when the sensors number can variate over time, which it is not the case. At least, right now, the current sensors number will be the only sensors addressed, and I see no problem in keeping it as it is. In the case of more sensor checks, or maybe setting flags on which sensors must bet monitoring/guard an extra logic can be addressed.

I think that we can finally merge and close this PR.

TimonSchreiber (approved):

PR #21, opened 2026-03-11
fix: send heartbeats in loop while waiting for GCS connection

TimonSchreiber:
MAVLinkController.waitGroundStation() previously sent a single heartbeat before timing out, causing the GCS handshake to fail if QGC wasn't immediately responsive. Changed to send heartbeats every 1000ms for up to 30 seconds per the MAVLink spec.

Additionally, MAVLinkService.launchService() no longer stops the service if no GCS responds within the timeout window — it logs a warning and continues. notifySystemReady() is now always called after waitSystemReady() to ensure telemetry data streams are unlocked regardless of system readiness state.

### DIFF
--- a/app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt
+++ b/app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt
@@ -34,6 +34,7 @@ import com.MAVLink.minimal.msg_heartbeat
 import io.getstream.log.taggedLogger
 import kotlin.math.roundToInt
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.delay
 import org.WenuLink.adapters.AircraftHandler
 import org.WenuLink.adapters.AsyncUtils
 import org.WenuLink.adapters.MessageRate
@@ -391,15 +392,22 @@ class MAVLinkController(
         )
     }
 
-    suspend fun waitGroundStation(timeout: Long = 5000L): Boolean {
+    suspend fun waitGroundStation(timeout: Long = 30000L): Boolean {
         if (!isTelemetryRunning()) return false
+
         // prevent to send data before initialization
         readOnlyMessageRate = true
-        // Send heartbeat and wait for any GCS
-        connectionController.sendHeartbeat(aircraft)
+
         logger.d { "Waiting for GCS." }
-        AsyncUtils.waitTimeout(10, timeout, ::isStationConnected)
+
+        // Send heartbeat and wait for any GCS
+        val deadline = System.currentTimeMillis() + timeout
+        while (!isStationConnected() && System.currentTimeMillis() < deadline) {
+            connectionController.sendHeartbeat(aircraft)
+            delay(1000L)
+        }
         logger.i { "GCS found?: ${isStationConnected()}." }
+
         return isStationConnected()
     }
 
--- a/app/src/main/java/org/WenuLink/mavlink/MAVLinkService.kt
+++ b/app/src/main/java/org/WenuLink/mavlink/MAVLinkService.kt
@@ -133,7 +133,8 @@ class MAVLinkService {
         if (hasParams) logger.d { "Parameters loaded" }
 
         isReady = controller.waitSystemReady(60000L)
-        if (isReady) controller.notifySystemReady()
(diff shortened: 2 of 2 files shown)


angel-ayala:
Hi Timon, I'm not sure about this PR , I also was facing troubles with the boot up / shutdown process, but the problem was more related to the coroutines lifecycles more than the app's logic.

The `MAVLinkController` class already ensures sending periodic messages. There is a list definition with all messages ID that should be sent, under a certain rate, initialized withe the `HEARTBEAT` message at 1Hz:

https://github.com/WenuLink/wenu-link-android/blob/c5d9afc43f710519918afada2ab2d5f58cfce354/app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt#L63-L68

Next, `MAVLinkService` class simultaneously registers the callback for incoming messages, and starts the periodic sending process:

https://github.com/WenuLink/wenu-link-android/blob/c5d9afc43f710519918afada2ab2d5f58cfce354/app/src/main/java/org/WenuLink/mavlink/MAVLinkService.kt#L109-L118

So basically, the `MAVLinkController.waitControlStation()` is outdated, no longer requiring the line 382:

https://github.com/WenuLink/wenu-link-android/blob/c5d9afc43f710519918afada2ab2d5f58cfce354/app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt#L377-L387

Additionally, your while loop proposed inside this function, do the same as `AsyncUtils.waitTimeout()` with the first argument as time delay in ms, and the second argument with the timeout in ms, and finally the function that will be called to check for the condition to be true.

Had you tried `feature/boot-sequence` 's code? If I'm not wrong in those changes a proper handling of start and stop sequence of the service has been addressed.
Since start takes a time to launch the services, the ActivityThread handles such function being called after the service is launched and before is stopped.
Additionally, when launched the WenuLinkService. if the Aircraft does not have a proper boot after a few seconds, no service should be executed.

angel-ayala:
I was looking up on the changed files, and you aimed to fix the bug with the last changes of `MAVLinkService`, by calling `controller.notifySystemReady()`, which partially solves it, since the `isReady` flag must be correctly set.
Therefore, the related bug here is not related to waiting for the GCS, but to waiting for the initial [communication flow](https://docs.qgroundcontrol.com/Stable_V5.0/en/qgc-dev-guide/communication_flow.html).
Basically, after sending a `HEARTBEAT`, in this case QGC, will first ask for the parameters and then for the mission items on the Aircraft. Right now, this is handled by setting the `wasRequested` attribute on `NavigationController` and `ParameterController` after a `MISSION_ACK` and `PARAM_REQUEST_LIST`, respectively. Nevertheless, the main issue here is on reconnecting after a successful connection.
Current thought logic waits for :

https://github.com/WenuLink/wenu-link-android/blob/c5d9afc43f710519918afada2ab2d5f58cfce354/app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt#L389-L395

And then mark the `isReady` flag in the `MAVLinkService` with the obtained results.

https://github.com/WenuLink/wenu-link-android/blob/c5d9afc43f710519918afada2ab2d5f58cfce354/app/src/main/java/org/WenuLink/mavlink/MAVLinkService.kt#L125-L126

In your case, you bypass the condition with `isReady` and directly calls `controller.notifySystemReady()`. Nevertheless, `isReady` is also required to be `true` for a full working `MAVLinkService`'s logic. Hence, the bug is related to the initial communication flow and the persistence of the connected status..
- Current implementation requires to know if parameters and mission were already sent to begin telemetry messages, due the large amount of messages required, avoiding network overload.
- After a successful connection, when restart the service, a proper waiting logic must considers events with sufficient timeout for `isReady = true`, while avoiding a time consuming lock.

Maybe more related to #13

TimonSchreiber:
Thank you for the detailed review @angel-ayala. After studying the feature/boot-sequence branch, I agree the coroutine lifecycle refactoring addresses the boot sequence more thoroughly than my patch did. The lone sendHeartbeat() call in waitGroundStation() is made redundant by launchSending() running concurrently, as you pointed out.
I'll close this PR and shift focus to testing feature/boot-sequence with our hardware setup.

PR #22, opened 2026-03-11
feature: code style and key management

TimonSchreiber:


angel-ayala:
Great!, this should be added before any other PR aiming to keep the styling on further commits. Can you also update the [CONTRIBUTING.md](https://github.com/WenuLink/wenu-link-android/blob/feature/boot-sequence/CONTRIBUTING.md) with a brief explanation, or if possible how to configure the IDE to adopt it?

TimonSchreiber:
Thanks for your quick reply @angel-ayala . I added ktlint to enforce the style guide, this should make working together a lot easier for all of us. Let me know if the explanation in CONTRIBUTING.md is enough and if the linter works for you.

app/build.gradle.kts
```diff
@@ -1,7 +1,31 @@
+import java.util.Properties
+
 plugins {
     alias(libs.plugins.android.application)
     alias(libs.plugins.kotlin.android)
     alias(libs.plugins.kotlin.compose)
+    id("org.jlleitschuh.gradle.ktlint") version "14.1.0"
```
angel-ayala: Is there any special reason for not following the `alias()` function as the others?
All libs should be included in the `libs.version.toml` with their corresponding version number.

app/build.gradle.kts
```diff
@@ -1,7 +1,31 @@
...
+}
+
+val localProps =
+    Properties().apply {
+        val f = rootProject.file("local.properties")
+        if (f.exists()) load(f.inputStream())
+    }
+
+ktlint {
+    version.set("1.8.0")
```
angel-ayala: Similarly but not completely sure, this value should come from `libs.versions.toml`. Possible the `coreKtx` key from `[versions]` tag? or `kotlin` if must point to the used version ?
If current tags does not represent such value, a new one must be created.

build.gradle.kts
```diff
@@ -3,4 +3,5 @@ plugins {
     alias(libs.plugins.android.application) apply false
     alias(libs.plugins.kotlin.android) apply false
     alias(libs.plugins.kotlin.compose) apply false
-}
\ No newline at end of file
+    id("org.jlleitschuh.gradle.ktlint") version "14.1.0"
```
angel-ayala: values must be declared once and used everywhere. Please look to previous comments referencing the `libs.versions.toml` file use.

angel-ayala (changes requested):
Great Job @TimonSchreiber !, I was wondering actually if the ending of a file should have and empty line or not. Only by observing the code, the styling seems pretty clear and useful to me. In addition to the separation of the API key, all changes were accepted.

The only things that must be addressed is the Lint library and its corresponding version that must be declared in the `libs.versions.toml` file.

Regards the Contributing.md brief and informative, nice that can be delegated to the IDE which should reduce out-of-style PRs.

angel-ayala:
I just cloned your branch and I'm unable to use the `./gradlew` command. When open the terminal from the Android Studio and run any command from the contributing.md I gets:

```text
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'WenuLink'.
> Could not resolve all artifacts for configuration 'classpath'.
   > Could not resolve com.android.tools.build:gradle:8.12.0.
     Required by:
         root project : > com.android.application:com.android.application.gradle.plugin:8.12.0
      > Dependency requires at least JVM runtime version 11. This build uses a Java 8 JVM.
```

I just updated the AGP  and same. I'm not sure why is saying `This build uses a Java 8 JVM.` I'm able to compile the APK and sync gradle with no problems. Possibly the terminal requires some `ENV` vars? Not sure what the problem is.

My test so far was, cloning your branch,  open AndroidStudio, and then:
- build -> Success
- delete the `{` and `}` of any `AircraftHandler`'s  if condition, my expectation on the IDE was to auto-fix, or only with gradlew can be done not sure what expects from AndroidStudio?

TimonSchreiber:
I had the same issue when setting up Android Studio for the first time.
To be able to run ./gradlew commands from the Android Studio terminal you need to tell it where the JDK is located that Android Studio uses for building, it can not use other JDKs you have installed on your machine. For that you need to add the Android Studio's bundled JDK to your system's PATH variable on windows and set the `JAVA_HOME` variable to point there as well. If not changed during installation, the JDK is probably found here:
On Windows: `C:\Program Files\Android\Android Studio\jbr`
On macOs: `/Applications/Android Studio.app/Contents/jbr/Contents/Home`
On Linux: `/opt/android-studio/jbr`

The auto-fix only works when the ktlint plugin is installed, not because ktlint was added to the Gradle Build.

Since this was an Android Studio problem for me, I didn't consider adding these steps to the contributing.md. If you want we can metion it under the setup section, for example:

```
#### Running Gradle commands

`./gradlew` commands must be run using the same JDK that Android Studio
uses for building. The easiest way is to use Android Studio's built-in
terminal, then set JAVA_HOME to point to Android Studio's bundled JDK.

The path varies by OS:
Windows: `C:\Program Files\Android\Android Studio\jbr`
macOS: `/Applications/Android Studio.app/Contents/jbr/Contents/Home`
Linux: `/opt/android-studio/jbr`
```

CONTRIBUTING.md
```diff
@@ -161,6 +165,37 @@ Regards class naming convention:
...
+
+    ./gradlew addKtlintCheckGitPreCommitHook
+
+This installs a local git hook that aborts the commit if any style
+violations are found. Run `./gradlew ktlintFormat` to fix them, then
+commit again. The hook is not committed to the repository, so each
+contributor who wants it must run this command once themselves.
+
+Optionally, install the [ktlint Android Studio plugin](https://plugins.jetbrains.com/plugin/15057-ktlint)
+for real-time violation highlighting in the editor.
```
angel-ayala, line 211:
```suggestion

#### Running Gradle commands

`./gradlew` commands must be run using the same JDK that Android Studio
uses for building. The easiest way is to use Android Studio's built-in
terminal, then set JAVA_HOME to point to Android Studio's bundled JDK.

The path varies by OS:
Windows: `C:\Program Files\Android\Android Studio\jbr`
macOS: `/Applications/Android Studio.app/Contents/jbr/Contents/Home`
Linux: `/opt/android-studio/jbr`

```

angel-ayala (changes requested):
exactly! I was able to run by executing `export JAVA_HOME=/opt/android-studio/jbr` on the terminal.
Everything seems to work fine, but I think is important to mention it on the Contributing.MD the proposed paragraph for that, should be enough.

angel-ayala (approved):
Great! Thanks @TimonSchreiber I will merge it, and all derived branches must be accordingly updated to enforce styling and maintainability.

PR #23, opened 2026-03-16
Refactor: MissionHandler readability improvements

TimonSchreiber:
Applies several small readability improvements to MissionHandler identified
during compatibility testing. No functional changes.

- updateState(): if-else chain → when expression with explicit else branch
  that preserves the implicit "no match = unchanged" behavior
- setSpeed(): coerceIn() with a named range; log after clamping so the
  message reflects what actually happened
- getItemCoordinates(): intermediate variables added no clarity, collapsed
  to expression body
- addWaypointNode(): NAV_WAYPOINT case extracted to private
  assembleWaypointNode() to reduce nesting in the when block


angel-ayala (approved):
Changes seems ok to me.
However, this should be refactored in a future for proper SDK classes `TimelineMission` and  `WaypointMissionOperator` swapping, for event-based (e.g. `MAV_CMD_DO_REPOSITION`) and mission-based (`MAV_CMD_MISSION_START`) commands.

Issue #25, opened 2026-03-22
Implement TelemetryController


angel-ayala:
There is no organization related to the available messages that can be sent using different `messagesRates`.
Hence, a new `controller` named `TelemetryController` is essential to keep things organized.
Currently, `MAVLinkController`'s variables named as `readOnlyMessageRate`, `messageRates`, and `availableDataList`, and methods `processDataStreamRequest`, `processMessageInterval`, and `setMessageRate`, must be moved keeping the current implemented logic into the new `TelemetryController`.

Changes aims to organize stream messages as declared in #7. It is essential to merge PR #17 before this.

PR #26, opened 2026-03-23
Feature/telemetry controller

TimonSchreiber:
## Summary

Implements #25.
Extracts all telemetry stream management out of
`MAVLinkController` into a dedicated `TelemetryController` class,
keeping the current logic intact.
## What moved

| Member | From | To |
|---|---|---|
| `readOnlyMessageRate` | `MAVLinkController` | `TelemetryController` |
| `messageRates` | `MAVLinkController` | `TelemetryController` |
| `availableDataList` | `MAVLinkController` | `TelemetryController` (private) |
| `setMessageRate()` | `MAVLinkController` | `TelemetryController` |
| `processDataStreamRequest()` | `MAVLinkController` | `TelemetryController` |
| `processMessageInterval()` | `MAVLinkController` | `TelemetryController` |
| `sendMessages()` | `MAVLinkController` | `TelemetryController` |

Direct field access to `readOnlyMessageRate` replaced by
`lockRates()`/`unlockRates()` methods on `TelemetryController`.

## Additional cleanup in MAVLinkController

- `client` changed from `MAVLinkClient?` to `lateinit var MAVLinkClient`,
  removing all `?.` safe calls
- Added typed computed properties `navigationController`,
  `parameterController`, and `telemetryController` consistent with the
  existing `connectionController` pattern, replacing inline
  `filterIsInstance` calls at usage sites
- Removed redundant `isEmpty()` guard before the `sendMessages` loop

## Merge note

This branch was developed on top of `feature/boot-sequence` and
**must not be merged before PR #17 is merged into `develop`**. Once
#17 merges, this branch should be rebased onto `develop` before
review is finalised.

### DIFF
--- a/app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt
+++ b/app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt
@@ -101,9 +101,8 @@ class MAVLinkController(private val aircraft: AircraftHandler) {
         when (commandMsg.command) {
             MAV_CMD.MAV_CMD_REQUEST_MESSAGE -> processRequestLong(commandMsg)
 
-            MAV_CMD.MAV_CMD_SET_MESSAGE_INTERVAL -> telemetryController.processMessageInterval(
-                commandMsg
-            )
+            MAV_CMD.MAV_CMD_SET_MESSAGE_INTERVAL ->
+                telemetryController.processMessageInterval(commandMsg)
 
             // TODO: Unhandled command ID: 521.
             // MAV_CMD.MAV_CMD_REQUEST_CAMERA_INFORMATION -> {}
@@ -187,14 +186,14 @@ class MAVLinkController(private val aircraft: AircraftHandler) {
     }
 
     fun notifySystemReady() {
-        telemetryController.unlockRates()
+        telemetryController.unlockSend()
     }
 
     fun isStationConnected(): Boolean = connectionController.isGCSPresent
 
     suspend fun waitGroundStation(timeout: Long = 5000L): Boolean {
         // prevent to send data before initialization
-        telemetryController.lockRates()
+        telemetryController.lockSend()
         // Wait for GCS heartbeat
         logger.d { "Waiting for GCS." }
         AsyncUtils.waitTimeout(10, timeout, ::isStationConnected)
--- a/app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt
+++ b/app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt
@@ -4,7 +4,6 @@ import com.MAVLink.Messages.MAVLinkMessage
 import com.MAVLink.common.msg_command_int
 import com.MAVLink.common.msg_command_long
 import com.MAVLink.common.msg_home_position
-import com.MAVLink.common.msg_request_data_stream
 import com.MAVLink.enums.MAV_CMD
 import io.getstream.log.taggedLogger
 import org.WenuLink.adapters.AsyncUtils
@@ -70,9 +69,6 @@ class MAVLinkController(private val aircraft: AircraftHandler) {
 
             msg_command_int.MAVLINK_MSG_ID_COMMAND_INT -> processCommandInt(msg)
 
-            msg_request_data_stream.MAVLINK_MSG_ID_REQUEST_DATA_STREAM ->
-                telemetryController.processDataStreamRequest(msg)
-
             else -> {
                 // Notify if no message ID definition was found in any Controller
                 logger.w { "Unhandled message ${msg.name()}" }
@@ -100,9 +96,6 @@ class MAVLinkController(private val aircraft: AircraftHandler) {
 
         when (commandMsg.command) {
             MAV_CMD.MAV_CMD_REQUEST_MESSAGE -> processRequestLong(commandMsg)
-
-            MAV_CMD.MAV_CMD_SET_MESSAGE_INTERVAL ->
-                telemetryController.processMessageInterval(commandMsg)
 
             // TODO: Unhandled command ID: 521.
             // MAV_CMD.MAV_CMD_REQUEST_CAMERA_INFORMATION -> {}
--- a/app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt
+++ b/app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt
@@ -1,43 +1,15 @@
 package org.WenuLink.controllers
 
 import com.MAVLink.Messages.MAVLinkMessage
-import com.MAVLink.common.msg_attitude
-import com.MAVLink.common.msg_battery_status
 import com.MAVLink.common.msg_command_int
 import com.MAVLink.common.msg_command_long
-import com.MAVLink.common.msg_control_system_state
-import com.MAVLink.common.msg_estimator_status
-import com.MAVLink.common.msg_gimbal_device_information
-import com.MAVLink.common.msg_global_position_int
-import com.MAVLink.common.msg_gps_raw_int
-import com.MAVLink.common.msg_gps_status
 import com.MAVLink.common.msg_home_position
-import com.MAVLink.common.msg_local_position_ned
-import com.MAVLink.common.msg_mag_cal_report
-import com.MAVLink.common.msg_mission_current
-import com.MAVLink.common.msg_nav_controller_output
-import com.MAVLink.common.msg_onboard_computer_status
-import com.MAVLink.common.msg_power_status
-import com.MAVLink.common.msg_radio_status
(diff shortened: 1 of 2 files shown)


angel-ayala:
Current changes are too much and most of them related to `feature/boot-sequence` once we merged with `develop`  I will do the review.

angel-ayala:
@TimonSchreiber can you merge develop to this branch, in that way specific changes from this PR only will be shown, thanks.

TimonSchreiber:
@angel-ayala you can review it now

app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt
```diff
@@ -198,13 +101,15 @@ class MAVLinkController(private val aircraft: AircraftHandler) {
         when (commandMsg.command) {
             MAV_CMD.MAV_CMD_REQUEST_MESSAGE -> processRequestLong(commandMsg)
 
-            MAV_CMD.MAV_CMD_SET_MESSAGE_INTERVAL -> processMessageInterval(commandMsg)
+            MAV_CMD.MAV_CMD_SET_MESSAGE_INTERVAL -> telemetryController.processMessageInterval(
+                commandMsg
+            )
```
angel-ayala:
```suggestion
            MAV_CMD.MAV_CMD_SET_MESSAGE_INTERVAL ->
                telemetryController.processMessageInterval(commandMsg)
```

app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt
```diff
@@ -271,110 +176,25 @@ class MAVLinkController(private val aircraft: AircraftHandler) {
...
-                commandMsg.command,
-                MAV_RESULT.MAV_RESULT_ACCEPTED
-            )
-        )
+        telemetryController.sendMessages(controllers, aircraft)
     }
 
     fun notifySystemReady() {
-        readOnlyMessageRate = false
+        telemetryController.unlockRates()
```
angel-ayala:
```suggestion
        telemetryController.unlockSend()
```

app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt
```diff
@@ -271,110 +176,25 @@ class MAVLinkController(private val aircraft: AircraftHandler) {
...
     }
 
     fun isStationConnected(): Boolean = connectionController.isGCSPresent
 
     suspend fun waitGroundStation(timeout: Long = 5000L): Boolean {
         // prevent to send data before initialization
-        readOnlyMessageRate = true
+        telemetryController.lockRates()
```
angel-ayala:
```suggestion
        telemetryController.lockSend()
```

app/src/main/java/org/WenuLink/controllers/TelemetryController.kt
```diff
@@ -0,0 +1,213 @@
...
+import kotlin.math.roundToInt
+import org.WenuLink.adapters.MessageUtils
+import org.WenuLink.adapters.aircraft.AircraftHandler
+import org.WenuLink.adapters.aircraft.MessageRate
+import org.WenuLink.mavlink.MAVLinkClient
+
+class TelemetryController(override val client: MAVLinkClient) : IController {
+    private val logger by taggedLogger(TelemetryController::class.java.simpleName)
+
+    private var readOnlyMessageRate = true
```
angel-ayala:
```suggestion
    private var setOnlyMessageRate = true
```

Seems better to be a setOnly flag to allows the `messageRates` being updated as required in any moment, but only allows to send at such rates when `setOnlyMessageRate` flag is `false`.

app/src/main/java/org/WenuLink/controllers/TelemetryController.kt
```diff
@@ -0,0 +1,213 @@
...
+            currentRate.microSecondsInterval = microSecondsInterval
+        }
+        return currentRate
+    }
+
+    @Synchronized
+    fun processDataStreamRequest(msg: MAVLinkMessage) {
+        // https://ardupilot.org/dev/docs/mavlink-requesting-data.html
+        // https://mavlink.io/en/messages/common.html#MAV_DATA_STREAM
+        if (readOnlyMessageRate) return
```
angel-ayala:
```suggestion
```

app/src/main/java/org/WenuLink/controllers/TelemetryController.kt
```diff
@@ -0,0 +1,213 @@
...
+                ((1.0 / timeInterval.toFloat()) * 1_000_000).roundToInt()
+            }
+        }
+
+        dataList.forEach { setMessageRate(it, timeInterval.toLong()) }
+    }
+
+    @Synchronized
+    fun processMessageInterval(commandMsg: msg_command_long) {
+        if (readOnlyMessageRate) return
```
angel-ayala:
```suggestion
```

app/src/main/java/org/WenuLink/controllers/TelemetryController.kt
```diff
@@ -0,0 +1,213 @@
...
+            MessageUtils.msgCommandAck(
+                commandMsg.command,
+                MAV_RESULT.MAV_RESULT_ACCEPTED
+            )
+        )
+    }
+
+    fun lockRates() {
+        readOnlyMessageRate = true
+    }
```
angel-ayala:
```suggestion
    fun lockSend() {
        setOnlyMessageRate = true
    }
```

app/src/main/java/org/WenuLink/controllers/TelemetryController.kt
```diff
@@ -0,0 +1,213 @@
...
+
+    fun unlockRates() {
+        readOnlyMessageRate = false
+    }
```
angel-ayala:
```suggestion
    fun unlockSend() {
        setOnlyMessageRate = false
    }
```

app/src/main/java/org/WenuLink/controllers/TelemetryController.kt
```diff
@@ -0,0 +1,213 @@
...
+
+    fun sendMessages(controllers: List<IController>, aircraft: AircraftHandler) {
+        if (!(client.mustProcessMessages())) {
+            logger.w { "MAVLink client is not ready!" }
+            return
+        }
+
+        val currentMicroTime = MessageUtils.getMicroTime()
+        for (rate in messageRates) {
+            if (rate.microSecondsInterval == -1L) continue // skip if deactivated
```
angel-ayala:
```suggestion
            if (setOnlyMessageRate && rate.MessageID != 0 ) continue // skip if setOnly mode for no HEARTBEAT messages
            if (rate.microSecondsInterval == -1L) continue // skip if deactivated
```

angel-ayala (changes requested):
Nice job @TimonSchreiber , all requested changes seems to be addressed and now is more compatible with the architecture.
A possible bug is the the set process of `messageRates` being blocking when receiving the `SET_MESSAGE_INTERVAL` message, instead of preventing to send any message if it was not unlocked to do.
I pointed out some suggestions alongside the files, but I'm not sure if those changes are sufficient to fulfill the expected behavior of: waiting until the parameters and mission items were requested to start sending any telemetry related message. Such behavior is expecting due to the communication flow start process.

TimonSchreiber:
@angel-ayala I implemented the requested changes and agree that this is the better approach at handling incoming `messageRates`. Would you consider `isTelemetrySuppressed`, `holdTelemetry` or something similar for `setOnlyMessageRate`? Neither `readOnly` nor `setOnly` are immediately intuitive to me in their intent.

angel-ayala:
Yes, makes sense to me in addressing a more proper name given the intent.
Both suggestions seems ok to me, but I would not use the Telemetry word on its definition, since is an attribute of the `TelemetryController` and we can assume that is telemetry related.
Since the intent is to prevent message sending, a name suggestion can be `supressBroadcast` or anything similar to represent the condition inside `sendMessages`. Any name change should also update the methods' name to properly reflect the changes, considering the same example, `lockSend` can be renamed to `stopBroadcast` and `unlockSend` to `startBroadcast`.

If you have any other naming idea that fits our discussion, feel free to use it.

angel-ayala:
Oh my bad, I just now realized that there is no method inheritance from `IController`, to keep the architecture choices, `TelemetryController` must inherit `processMessage` and `processCommandLong` with their respective messages and being called when iterate the `controllers` list on `MAVLinkController`.

I think that should be enough to maintain the code structure and style

TimonSchreiber:
Good catch @angel-ayala I totally missed that. This should now work as intended.

angel-ayala (approved):
Well done @TimonSchreiber , everything seems to be ok now.

PR #27, opened 2026-03-24
Add camera service capability

angel-ayala:
The following changes aims to address #14 by implementing the Camera protocol v2 as stated in [MAVLink documentation](https://mavlink.io/en/services/camera.html). Changes has already merged from `feature/boot-sequence` to include recent architectural changes for the init/shutdown sequence. Nevertheless, here I'm also using the Command pattern and I was thinking in implement an abstraction of such logic, since it will be used with different async actions.
Additionally, off-topic, `MissionHandler` must be updated to use the Command pattern design too.

However, such changes to refactor the Command pattern in an abstract reusable way I think is better to apply under a different issue/PR.
Right now, maybe we could test if the current logic is ok before move into a more appropriate architecture.

Revisions also should wait for #17  to be merged with `develop`


app/src/main/java/org/WenuLink/sdk/CameraManager.kt
```diff
@@ -103,4 +213,143 @@ object CameraManager {
     }
 
     fun isCodecStarted(): Boolean = codecManager != null
+
+    private suspend fun setCameraMode(mode: SettingsDefinitions.CameraMode): String? =
+        suspendCancellableCoroutine { cont ->
+            if (cameraMode == mode) {
+                cont.resume(null)
```
TimonSchreiber, line 220:
```suggestion
                cont.resume(null)
                return@suspendCancellableCoroutine
```

app/src/main/java/org/WenuLink/sdk/CameraManager.kt
```diff
@@ -45,6 +61,100 @@ object CameraManager {
...
+            override fun onFailure(p0: DJIError?) {
+                if (p0 != null) {
+                    onResult(p0.description, false)
+                }
+            }
+        }
+
+    suspend fun retrieveFirmwareVersion(): String? = suspendCancellableCoroutine { cont ->
+        if (fwVersion != null) {
+            cont.resume(fwVersion)
```
TimonSchreiber, line 79:
```suggestion
            cont.resume(fwVersion)
            return@suspendCancellableCoroutine
```

app/src/main/java/org/WenuLink/sdk/CameraManager.kt
```diff
@@ -45,6 +61,100 @@ object CameraManager {
...
+        mInstance?.getFirmwareVersion(
+            createCompletionCallback { firmwareVersion, _ ->
+                cont.resume(firmwareVersion)
+            }
+        ) ?: cont.resume(null)
+    }
+
+    suspend fun retrieveSerialNumber(): String? = suspendCancellableCoroutine { cont ->
+        if (serialNumber != null) {
+            cont.resume(serialNumber)
```
TimonSchreiber, line 91:
```suggestion
            cont.resume(serialNumber)
            return@suspendCancellableCoroutine
```

app/src/main/java/org/WenuLink/controllers/CameraController.kt
```diff
@@ -0,0 +1,442 @@
...
+            lon = mavData.longitude
+            alt = mavData.altitude
+            relative_alt = mavData.relativeAltitude
+            q = OrientationUtils.eulerDegToQuaternion(
+                rollDeg = mavData.roll.toDouble(),
+                pitchDeg = mavData.pitch.toDouble(),
+                yawDeg = mavData.yaw.toDouble()
+            ).toFloatArray()
+            image_index = imageInfo.index
+            capture_result = imageInfo.captureOk.toString().toByte()
```
TimonSchreiber:
```suggestion
            capture_result = (if (imageInfo.captureOk) 1 else 0).toByte()
```

app/src/main/java/org/WenuLink/controllers/CameraController.kt
```diff
@@ -0,0 +1,442 @@
...
+            focal_length = Float.NaN
+            sensor_size_h = Float.NaN
+            sensor_size_v = Float.NaN
+            resolution_h = cameraInfo.width
+            resolution_v = cameraInfo.height
+            lens_id = 0
+            flags = CAMERA_CAP_FLAGS.CAMERA_CAP_FLAGS_CAPTURE_IMAGE.toLong() or
+                CAMERA_CAP_FLAGS.CAMERA_CAP_FLAGS_CAPTURE_VIDEO.toLong() or
+                CAMERA_CAP_FLAGS.CAMERA_CAP_FLAGS_HAS_MODES.toLong()
+            cam_definition_version = cameraInfo.fwVersion.toInt()
```
TimonSchreiber: `cam_definition_version` does not store the firmware version but the camera definition version, a different value. If unknown it should be "0" per documentation.

app/src/main/java/org/WenuLink/controllers/CameraController.kt
```diff
@@ -0,0 +1,442 @@
...
+                    )
+                }
+            }
+    }
+
+    fun msgCameraInformation(cameraInfo: CameraMetadata): msg_camera_information =
+        msg_camera_information().apply {
+            vendor_name = MessageUtils.toShortArray("DJI", 32)
+            model_name = MessageUtils.toShortArray(cameraInfo.name, 32)
+            firmware_version = MessageUtils.packVersion(
```
TimonSchreiber: This should access the `CameraMetadata.fwVersion` instead of a hard coded value.
Since `fwVersion` is not a plain String but a value like "01.03.0600", this would need to get parsed before passing to `MessageUtils.packVersion()`

app/src/main/java/org/WenuLink/adapters/camera/CameraHandler.kt
```diff
@@ -0,0 +1,164 @@
...
+            )
+        )
+        logger.d { "availableCameras(${availableCameras.size})" }
+        setMode(CAMERA_MODE.CAMERA_MODE_IMAGE, 0)
+
+        logger.d { "Managing ${availableCameras.size} detected camera(s)" }
+        return true
+    }
+
+    fun getCamera(cameraIdx: Int): CameraMetadata? = availableCameras.getOrNull(cameraIdx)
```
TimonSchreiber:
```suggestion
    fun getCamera(cameraIdx: Int): CameraMetadata? = availableCameras.firstOrNull { it.id == cameraId }
```

app/src/main/java/org/WenuLink/adapters/camera/CameraHandler.kt
```diff
@@ -0,0 +1,164 @@
...
+        cameraIdx
+    )
+
+    fun captureIdle(cameraIdx: Int = 0): Boolean = checkCaptureStatus(
+        CameraCaptureStatus.IDLE,
+        cameraIdx
+    )
+
+    fun isPhotoMode(cameraIdx: Int = 0): Boolean =
+        (getCamera(cameraIdx)?.state?.mavlinkMode ?: -1) == CAMERA_MODE.CAMERA_MODE_IMAGE
```
TimonSchreiber:
```suggestion
        getCamera(cameraIdx)?.state?.mavlinkMode == CAMERA_MODE.CAMERA_MODE_IMAGE
```

app/src/main/java/org/WenuLink/adapters/camera/CameraHandler.kt
```diff
@@ -0,0 +1,164 @@
...
+
+    fun isVideoMode(cameraIdx: Int = 0): Boolean =
+        (getCamera(cameraIdx)?.state?.mavlinkMode ?: -1) == CAMERA_MODE.CAMERA_MODE_VIDEO
```
TimonSchreiber:
```suggestion
        getCamera(cameraIdx)?.state?.mavlinkMode == CAMERA_MODE.CAMERA_MODE_VIDEO
```

app/src/main/java/org/WenuLink/adapters/camera/CameraHandler.kt
```diff
@@ -0,0 +1,164 @@
...
+                    state = it.state.copy(
+                        captureStatus = newStatus
+                    )
+                )
+            )
+        }
+    }
+
+    fun checkCaptureStatus(status: CameraCaptureStatus, cameraIdx: Int = 0): Boolean =
+        availableCameras[cameraIdx].state.captureStatus == status
```
TimonSchreiber:
```suggestion
        getCamera(cameraIdx)?.state?.captureStatus == status
```

app/src/main/java/org/WenuLink/controllers/MAVLinkController.kt
```diff
@@ -342,7 +343,7 @@ class MAVLinkController(private val aircraft: AircraftHandler) {
                 1_000_000
             } // Hz to micro seconds if must start
             else {
-                ((1.0 / timeInterval.toFloat()) * 1_000_000).roundToInt()
+                ((1.0 / timeInterval.toFloat()) * 1_000).roundToInt()
```
TimonSchreiber: Are you sure about this change? If yes please adjust the comment since this no longer converts Hz to micro seconds.

app/src/main/java/org/WenuLink/controllers/CameraController.kt
```diff
@@ -0,0 +1,442 @@
...
+        }
+        return cameraInfo
+    }
+
+    private fun requestShootPhoto(
+        commandLongMsg: msg_command_long,
+        aircraft: AircraftHandler,
+        cameraInfo: CameraMetadata
+    ) {
+        val intervalTime: Float = commandLongMsg.param2 // seconds
```
TimonSchreiber:
```suggestion
        val intervalTime: Long = (commandLongMsg.param2 * 1_000).toLong() // milliseconds
```

app/src/main/java/org/WenuLink/controllers/CameraController.kt
```diff
@@ -0,0 +1,442 @@
...
+            aircraft.cameras.sequenceIndex != -1
+        } else {
+            aircraft.cameras.sequenceIndex <= totalPhotos
+        }
+
+        aircraft.cameras.sequenceIndex = initSequence
+
+        while (mustCapture()) {
+            if ((System.currentTimeMillis() - aircraft.cameras.captureTimestamp) < intervalTime) {
+                continue
```
TimonSchreiber, line 213:
```suggestion
                delay(100)
                continue
```

app/src/main/java/org/WenuLink/controllers/CameraController.kt
```diff
@@ -0,0 +1,442 @@
...
+            }
+            if (!aircraft.cameras.captureIdle(cameraInfo.id)) {
+                continue
```
TimonSchreiber, line 216:
```suggestion
                delay(100)
                continue
```

TimonSchreiber (changes requested):
Nice work so far! I have a few comments before we merge:

### Bugs:

#### 1. `CameraManager`: Missing `return` after early `const.resume()`
`setCameraMode`, `retrieveFirmwareVersion` and `retrieveSerialNumber` all resume the continuation early but then fall through to the SDK call, which will attempt a second `resume`.

#### 2. `CameraController`
`capture_result` encoding throws `NumberFormatException`
```kotlin
// Current: "true".toByte() throws NumberFormatException
capture_result = imageInfo.captureOk.toString().toByte()

// Fix
capture_result = (if (imageInfo.captureOk) 1 else 0).toByte()
```

#### 3. `CameraController`
`requestShootPhoto` needs a delay to not run countless times on the early returns. Also compares milliseconds from `System.currentTimeMillis()` with seconds from `intervalTime`.

```kotlin
val intervalTime: Long = (commandLongMsg.param2 * 1000).toLong() // milliseconds
// ...
while (mustCapture()) {
    if ((System.currentTimeMillis() - aircraft.cameras.captureTimestamp) < intervalTime) {
        delay(100)  // add this
        continue
    }
    if (!aircraft.cameras.captureIdle(cameraInfo.id)) {
        delay(100)  // add this
        continue
    }
    // ...
}
```

#### 4. `CameraController` and `CameraHandler`: Camera ID is 1-based, list is 0-based

The MAVLink spec defines camera IDs as 1-based (param value `1` = first camera). `CameraHandler` initialises the single camera with `id = 1` at list index `0`. But `getCamera()` currently uses the ID directly as a list index:

```kotlin
// Current: getCamera(1) returns availableCameras[1] → null (out of bounds)
fun getCamera(cameraIdx: Int): CameraMetadata? = availableCameras.getOrNull(cameraIdx)
```

When QGC sends a video start/stop or capture command targeting camera `1`, the lookup returns `null` and the command is silently rejected. Fix by looking up by ID instead of index:

```kotlin
// Fix
fun getCamera(cameraId: Int): CameraMetadata? =
    availableCameras.firstOrNull { it.id == cameraId }
```

Also extend the use of `getCamera()` to the `checkCaptureStatus()` to avoid `IndexOutOfBoundsException`

---

### Design Issues

#### 5. `CameraController.msgCameraInformation()` uses hardcoded firmware version:
The infrastructure to populate this already exists: `CameraMetadata.fwVersion` is populated from `CameraManager.retrieveMetadata()`. The DJI firmware string (e.g. `"01.03.0600"`) would need to be parsed before passing into `packVersion`. Suggested approach:

```kotlin
// Parse DJI fw string "MM.mm.pppp" into major/minor/patch
val parts = cameraInfo.fwVersion.split(".").mapNotNull { it.toIntOrNull() }
firmware_version = if (parts.size >= 3) {
    MessageUtils.packVersion(parts[0], parts[1], parts[2], FIRMWARE_VERSION_TYPE.FIRMWARE_VERSION_TYPE_OFFICIAL)
} else {
    MessageUtils.packVersion(0, 0, 0, FIRMWARE_VERSION_TYPE.FIRMWARE_VERSION_TYPE_DEV)
}
```

Also, `cam_definition_version = cameraInfo.fwVersion.toInt()` will throw for the same reason, `"01.03.0600"` is not a plain integer. This field is meant to version the *camera definition XML file*, not the firmware. It should be a simple hardcoded constant '0' until a definition file is actually hosted.

#### 6. `CameraManager.kt`: `createCompletionCallback` silently hangs on null error

```kotlin
override fun onFailure(p0: DJIError?) {
    if (p0 != null) {
        onResult(p0.description, false)
    }
    // if p0 IS null → onResult never called → coroutine hangs indefinitely
}
```

Fix:
```kotlin
override fun onFailure(p0: DJIError?) {
    onResult(p0?.description ?: "Unknown error", false)
}
```

#### 7. `CameraHandler.kt`: Redundant polling after `retrieveMetadata()`
`initCameras` calls `CameraManager.retrieveMetadata()` (which is itself a suspend function that already awaits both `fwVersion` and `serialNumber`), and then immediately polls for those same fields with `AsyncUtils.waitTimeout`. By the time `retrieveMetadata()` returns they are already set. Either trust the await or replace `retrieveMetadata()` with only the polling loops, not both.

#### 8. `MAVLinkController.kt`: Time interval unit change needs verification

```kotlin
// Before
((1.0 / timeInterval.toFloat()) * 1_000_000).roundToInt()
// After
((1.0 / timeInterval.toFloat()) * 1_000).roundToInt()
```

The adjacent comment reads "Hz to micro seconds", which would imply `* 1_000_000` is correct. If the downstream consumer actually expects milliseconds the new value is right, but this should be explicitly verified against wherever the resulting value is consumed before accepting.

#### 9. `CameraController.kt`: `msgCameraInformation` capability flags are hardcoded

```kotlin
flags = CAMERA_CAP_FLAGS_CAPTURE_IMAGE or CAMERA_CAP_FLAGS_CAPTURE_VIDEO or CAMERA_CAP_FLAGS_HAS_MODES
```

`CAMERA_CAP_FLAGS_HAS_VIDEO_STREAM` is absent. If the camera actually streams video (which it does via WebRTC), this flag should be set too so QGC offers the stream UI.

angel-ayala:
That was really good review @TimonSchreiber , indeed I was noticing and error with "01.03.0600" on the logcat but I didn't dig deeper to understand what was going on. I addressed all the changes with exception Bug: 3 where you suggests to add a `delay()` function. For adding such kind of function, the method must be suspendable, changing all declarations related to it, required for addressing that change. I understand your concern, but I see no harm in keeping it as it is right now, since it will not dispatch the `TakePhotoCommand` until conditions are met.

A proper way to address this, could be by using [MissionActions](https://developer.dji.com/api-reference/android-api/Components/Missions/DJIMissionAction.html) and corresponding listeners to react every time a photo is taken if possible. In any case, adding the changes from #28 should make easy the Command pattern use to address such execution with a proper suspendable method.

TimonSchreiber (dismissed):
While I can agree that the busy-wait for a singel photo is negligible, when it comes to multi-photo requests, that would mean 100% of thread utilization for the whole interval between shots. I am sure this can wait until #28, but the changes would be small in my opinion.
Just capture the `CameraHandler`'s `CoroutineScope` in a field and access it in the `requestStartCapture` via the `AircraftHandler.cameras` to launch the `requestShootPhoto` call:

```kotlin
private fun requestStartCapture(commandLongMsg: msg_command_long, aircraft: AircraftHandler) {
    // ... camera lookup and ack as before ...
    
    aircraft.cameras.launchCameraJob {
        requestShootPhoto(commandLongMsg, aircraft, cameraInfo)
    }
}
```

and in `CamerHandler`:

```kotlin
fun launchCameraJob(block: suspend () -> Unit): Job? =
    handlerScope?.launch { block() }
```

---

I also noticed the cyclical dependency between `MAVLinkController` and `CameraController` were both own an instance of each other. Since `CameraController` only needs to access the `MAVLinkController`'s `setMessageRate` method, maybe injecting only that one as a lambda would be enough:

```kotlin
class CameraController(
    override val client: MAVLinkClient,
    private val onSetMessageRate: (messageId: Int, intervalMs: Long) -> Unit
) : IController
```

and then

```kotlin
controllers += CameraController(client) { msgId, rate ->
    setMessageRate(msgId, rate)
}
```

I'd lean towards the lambda approach, but happy to leave it for a follow-up if you'd rather not touch the constructor now.

TimonSchreiber:
Please run `gradlew ktlintFormat` before merging.

angel-ayala:
my bad, I just addressed the changes and also added a missing command message that can be received.

I think that at this point is ok. Once the command PR is merged, I will rebase from develop and merge it. From that point, I will open a new PR to address the camera command pattern.

angel-ayala:
looking on files changed, I noticed that I pushed changes not related to the camera service. Instead, changes in `TelemetryData.kt` were focused to ensure the data transformation between MAVLink and SDK, considering different units scale and structured data. If I'm not wrong there is already being used in few moments, without entirely replacing the telemetry information parse for corresponding MAVLink messages. If is not ok for you I can do another commit removing it.

TimonSchreiber (approved):
This is ready for merge.

PR #28, opened 2026-03-25
Feature/command processor

angel-ayala:
These changes aims to extract the Command pattern processing logic contained in `AircraftHandler` in a higher level of abstraction to correctly reuse with other `IHandler`s classes.

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
+++ b/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
@@ -50,7 +50,7 @@ data class ArmCommand(val timeout: Long = 5000L) : AircraftCommand {
     }
 
     override suspend fun onStop(ctx: AircraftHandler) {
-        ctx.dispatchCommand(DisarmCommand()) { }
+        ctx.dispatchCommand(DisarmCommand())
     }
 }
 
@@ -83,7 +83,7 @@ data class TakeoffCommand(val initialAltitude: Float = 2f) : AircraftCommand {
         val isFlying = ctx.awaitFlightState(true)
 
         if (!isFlying) {
-            ctx.dispatchCommand(LandCommand(true)) { }
+            ctx.dispatchCommand(LandCommand(true))
             return "Unable to takeoff"
         }
 
@@ -91,7 +91,7 @@ data class TakeoffCommand(val initialAltitude: Float = 2f) : AircraftCommand {
     }
 
     override suspend fun onStop(ctx: AircraftHandler) {
-        ctx.dispatchCommand(LandCommand(true)) { }
+        ctx.dispatchCommand(LandCommand(true))
     }
 }
 
@@ -107,7 +107,7 @@ data class LandCommand(val withLandingConfirmation: Boolean = true) : AircraftCo
 
         if (!onTheGround) return "Unable to land"
 
-        ctx.dispatchCommand(DisarmCommand()) { }
+        ctx.dispatchCommand(DisarmCommand())
 
         return null
     }
--- a/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
+++ b/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
@@ -1,9 +1,11 @@
 package org.WenuLink.adapters.aircraft
 
-sealed interface AircraftCommand {
-    fun validate(ctx: AircraftHandler): String?
-    suspend fun execute(ctx: AircraftHandler): String?
-    suspend fun onStop(ctx: AircraftHandler)
+import org.WenuLink.adapters.commands.ICommand
+
+sealed interface AircraftCommand : ICommand<AircraftHandler> {
+    override fun validate(ctx: AircraftHandler): String?
+    override suspend fun execute(ctx: AircraftHandler): String?
+    override suspend fun onStop(ctx: AircraftHandler)
 }
 
 data class BootCommand(val timeout: Long = 5000L) : AircraftCommand {
@@ -48,7 +50,7 @@ data class ArmCommand(val timeout: Long = 5000L) : AircraftCommand {
     }
 
     override suspend fun onStop(ctx: AircraftHandler) {
-        ctx.dispatchCommand(DisarmCommand())
+        ctx.dispatchCommand(DisarmCommand()) { }
     }
 }
 
@@ -81,7 +83,7 @@ data class TakeoffCommand(val initialAltitude: Float = 2f) : AircraftCommand {
         val isFlying = ctx.awaitFlightState(true)
 
         if (!isFlying) {
-            ctx.dispatchCommand(LandCommand(true))
+            ctx.dispatchCommand(LandCommand(true)) { }
             return "Unable to takeoff"
         }
 
@@ -89,7 +91,7 @@ data class TakeoffCommand(val initialAltitude: Float = 2f) : AircraftCommand {
     }
 
     override suspend fun onStop(ctx: AircraftHandler) {
-        ctx.dispatchCommand(LandCommand(true))
+        ctx.dispatchCommand(LandCommand(true)) { }
     }
 }
 
@@ -105,7 +107,7 @@ data class LandCommand(val withLandingConfirmation: Boolean = true) : AircraftCo
 
         if (!onTheGround) return "Unable to land"
 
-        ctx.dispatchCommand(DisarmCommand())
+        ctx.dispatchCommand(DisarmCommand()) { }
 
         return null
     }
--- a/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
+++ b/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
@@ -2,23 +2,22 @@ package org.WenuLink.adapters.aircraft
 
 import com.MAVLink.enums.MAV_MODE_FLAG
 import io.getstream.log.taggedLogger
-import kotlin.coroutines.cancellation.CancellationException
 import kotlin.math.roundToLong
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.async
-import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import org.WenuLink.adapters.AsyncUtils
 import org.WenuLink.adapters.MissionHandler
+import org.WenuLink.adapters.commands.CommandProcessor
+import org.WenuLink.adapters.commands.ICommand
+import org.WenuLink.adapters.commands.IHandler
 import org.WenuLink.sdk.FCManager
 import org.WenuLink.sdk.MissionActionManager
 import org.WenuLink.sdk.MissionManager
 
-class AircraftHandler {
+class AircraftHandler : IHandler<AircraftHandler> {
     companion object {
         private var mInstance: AircraftHandler? = null
 
@@ -27,7 +26,7 @@ class AircraftHandler {
                 mInstance = AircraftHandler()
             }
             if (serviceScope != null) {
-                mInstance!!.registerHandlerScope(serviceScope)
+                mInstance!!.registerScope(serviceScope)
             }
 
             return mInstance!!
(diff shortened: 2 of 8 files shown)


app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
```diff
@@ -48,7 +50,7 @@ data class ArmCommand(val timeout: Long = 5000L) : AircraftCommand {
     }
 
     override suspend fun onStop(ctx: AircraftHandler) {
-        ctx.dispatchCommand(DisarmCommand())
+        ctx.dispatchCommand(DisarmCommand()) { }
```
TimonSchreiber:
```suggestion
        ctx.dispatchCommand(DisarmCommand())
```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
```diff
@@ -81,15 +83,15 @@ data class TakeoffCommand(val initialAltitude: Float = 2f) : AircraftCommand {
         val isFlying = ctx.awaitFlightState(true)
 
         if (!isFlying) {
-            ctx.dispatchCommand(LandCommand(true))
+            ctx.dispatchCommand(LandCommand(true)) { }
```
TimonSchreiber:
```suggestion
            ctx.dispatchCommand(LandCommand(true))
```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
```diff
@@ -81,15 +83,15 @@ data class TakeoffCommand(val initialAltitude: Float = 2f) : AircraftCommand {
...
             return "Unable to takeoff"
         }
 
         return null
     }
 
     override suspend fun onStop(ctx: AircraftHandler) {
-        ctx.dispatchCommand(LandCommand(true))
+        ctx.dispatchCommand(LandCommand(true)) { }
```
TimonSchreiber:
```suggestion
        ctx.dispatchCommand(LandCommand(true))
```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
```diff
@@ -105,7 +107,7 @@ data class LandCommand(val withLandingConfirmation: Boolean = true) : AircraftCo
 
         if (!onTheGround) return "Unable to land"
 
-        ctx.dispatchCommand(DisarmCommand())
+        ctx.dispatchCommand(DisarmCommand()) { }
```
TimonSchreiber:
```suggestion
        ctx.dispatchCommand(DisarmCommand())
```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
```diff
@@ -50,68 +49,18 @@ class AircraftHandler {
...
-    private var commandJob: Job? = null
-    private var currentCommandJob: Deferred<String?>? = null
-    private val commandChannel =
-        Channel<Pair<AircraftCommand, (String?) -> Unit>>(capacity = Channel.UNLIMITED)
+    private val processor = CommandProcessor(this)
     private var monitorJob: Job? = null
 
     init {
-        dispatchCommand(BootCommand(5000))
+        dispatchCommand(BootCommand(5000)) { }
```
```suggestion
```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
```diff
@@ -213,9 +162,9 @@ class AircraftHandler {
 
     private fun safetyChecks() {
         // --- Safety hooks ---
-        rcInput?.hasCenteredJoystick()?.let {
+        if (telemetry.hasRCInput()) {
```
```suggestion
        if (telemetry.hasActiveJoystickInput()) {
```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
```diff
@@ -350,11 +299,11 @@ class AircraftHandler {
...
-    fun registerHandlerScope(handlerScope: CoroutineScope) {
-        telemetry.registerHandlerScope(handlerScope)
-        registerMissionListeners(handlerScope)
-        startCommandProcessor(handlerScope)
-        startMonitorJob(handlerScope)
+    override fun registerScope(scope: CoroutineScope) {
+        telemetry.registerHandlerScope(scope)
+        registerMissionListeners(scope)
+        startMonitorJob(scope)
+        processor.start(scope)
```
TimonSchreiber, line 299:
```suggestion
        processor.start(scope)
        dispatchCommand(BootCommand(5000))
```

app/src/main/java/org/WenuLink/adapters/aircraft/TelemetryHandler.kt
```diff
@@ -289,4 +289,6 @@ class TelemetryHandler {
     } else {
         AircraftManager.getAirlinkData()
     }
+
+    fun hasRCInput(): Boolean = getRCData()?.hasCenteredJoystick() == false
```
TimonSchreiber:
```suggestion
    fun hasActiveJoystickInput(): Boolean = getRCData()?.hasCenteredJoystick() == false
```

app/src/main/java/org/WenuLink/adapters/commands/CommandProcessor.kt
```diff
@@ -0,0 +1,68 @@
...
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.launch
+
+class CommandProcessor<T : IHandler<T>>(private val handler: T) {
+    private var requestJob: Job? = null
+    private var commandJob: Deferred<String?>? = null
+    private val commandChannel =
+        Channel<Pair<ICommand<T>, (String?) -> Unit>>(Channel.UNLIMITED)
+
+
```
TimonSchreiber:
```suggestion
```

app/src/main/java/org/WenuLink/adapters/commands/IHandler.kt
```diff
@@ -0,0 +1,12 @@
+package org.WenuLink.adapters.commands
+
+import kotlinx.coroutines.CoroutineScope
+
+/**
+ * Interface class to implement the Command pattern through Job Handlers
+ */
+interface IHandler<T : IHandler<T>> {
+    fun registerScope(scope: CoroutineScope)
+    fun dispatchCommand(cmd: ICommand<T>, onResult: (String?) -> Unit)
```
TimonSchreiber:
```suggestion
    fun dispatchCommand(cmd: ICommand<T>, onResult: (String?) -> Unit = { })
```

app/src/main/java/org/WenuLink/controllers/NavigationController.kt
```diff
@@ -301,7 +301,7 @@ class NavigationController(override val client: MAVLinkClient) : IController {
                 coordinate,
                 if (commandIntMsg.param1 != -1f) commandIntMsg.param1 else null
             )
-        )
+        ) { error -> }
```
TimonSchreiber:
```suggestion
        )
```

app/src/main/java/org/WenuLink/adapters/commands/CommandProcessor.kt
```diff
@@ -0,0 +1,68 @@
...
+import kotlin.coroutines.cancellation.CancellationException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.async
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.launch
+
+class CommandProcessor<T : IHandler<T>>(private val handler: T) {
```
TimonSchreiber:
```suggestion
class CommandProcessor<T : IHandler<T>>(private val handler: T) {
    private val logger by taggedLogger(CommandProcessor::class.java.simpleName)
```

app/src/main/java/org/WenuLink/adapters/commands/CommandProcessor.kt
```diff
@@ -0,0 +1,68 @@
+package org.WenuLink.adapters.commands
+
```
TimonSchreiber, line 2:
```suggestion

import io.getstream.log.taggedLogger
```

app/src/main/java/org/WenuLink/adapters/commands/CommandProcessor.kt
```diff
@@ -0,0 +1,68 @@
...
+    private val commandChannel =
+        Channel<Pair<ICommand<T>, (String?) -> Unit>>(Channel.UNLIMITED)
+
+
+    fun start(scope: CoroutineScope) {
+        requestJob?.cancel()
+
+        requestJob = scope.launch {
+            for ((cmd, onResult) in commandChannel) {
+                try {
```
TimonSchreiber, line 25:
```suggestion
                try {
                    logger.d { "Executing: ${cmd::class.simpleName}" }

```

app/src/main/java/org/WenuLink/adapters/commands/CommandProcessor.kt
```diff
@@ -0,0 +1,68 @@
...
+
+                    val deferred = async {
+                        cmd.execute(handler)
+                    }
+
+                    commandJob = deferred
+
+                    val result = deferred.await()
+                    onResult(result)
+                } catch (e: CancellationException) {
```
TimonSchreiber, line 40:
```suggestion
                } catch (e: CancellationException) {
                    logger.w { "Command cancelled: ${cmd::class.simpleName}: $e" }
```

app/src/main/java/org/WenuLink/adapters/commands/CommandProcessor.kt
```diff
@@ -0,0 +1,68 @@
...
+                    cmd.onStop(handler)
+                    onResult(null)
+                } catch (e: Exception) {
```
TimonSchreiber, line 44:
```suggestion
                } catch (e: Exception) {
                    logger.e { "Command failed: ${cmd::class.simpleName} -> ${e.message}" }
```

app/src/main/java/org/WenuLink/adapters/commands/CommandProcessor.kt
```diff
@@ -0,0 +1,68 @@
...
+                    commandJob = null
+                }
+            }
+        }
+    }
+
+    fun dispatch(cmd: ICommand<T>, onResult: (String?) -> Unit = {}) {
+        val result = commandChannel.trySend(cmd to onResult)
+        if (result.isFailure) {
+            onResult("Failed to enqueue command")
```
TimonSchreiber:
```suggestion
            onResult("Failed to enqueue command: ${cmd::class.simpleName}")
```

TimonSchreiber (changes requested):
This will be a great architectural improvement!

I marked a few things:

### 1. Lost logging
The original `startCommandProcessor` logged three things that `CommandProcessor` drops:
* `logger.d { "Executing: ${cmd::class.simpleName}" }`
* `logger.w { "Command cancelled: ..." }`
* `logger.e { "Command failed: ..." }`

### 2. `hasRCInput()` is a misleading name
"hasRCInput" reads as "RC data is present", but what we want to express is more like "joystick is actively deflected". My suggestion would be `hasActiveJoystickInput()` or `isManualControlActive()`.

### 3. Default parameter dropped from interface
The old `dispatchCommand` has `onResult: (String?) - Unit = { }` as a default. `IHandler` dropps the default and now all `dispatchCommand` calls need the explicit `{ }` or `{ error -> }`. Adding the default back would be simple and make the rest easier to read.

### 4. Error message lost the class name
Old `"Failed to enqueue command: ${cmd::class.simpleName}"` is better than the new `"Failed to enqueue command"`

### 5. Dependency order
`BootCommand` is dispatched in `AircraftHandler.init` before `registerScope` is called, meaning it sits in the `UNLIMITED` channel until the processor starts. This works now since a handler is not operational until it has a scope, but `init` should only initialize state and not trigger behaviour that requires infrastructure that doesn't exist yet. A better approach would be to move the `BootCommand` dispatch out of `init` and into `registerScope` after `process.start(scope)`:

```kotlin
override fun registerScope(scope: CoroutineScope) {
    telemetry.registerHandlerScope(scope)
    registerMissionListeners(scope)
    startMonitorJob(scope)
    processor.start(scope)
    dispatchCommand(BootCommand(5000)) { }  // processor is ready now
}
```

---

I think it might be a good idea to merge this pr first and then rebase develop onto #27 so that we get the improved architecture for the `CameraHandler` and `CameraController` as well

---

Please make sure your linter setup works correctly. While only a minor formatting issue, the ktlint should have cought it either in the preCommitHook or using `ktlinFormat`. I had noticed issues in the previous pr as well, but didn't want to block it because of that since the next commit would have cought it anyway. But since it happend again, it is safe to assume your setup didn't work correctly.

angel-ayala:
I just addressed your comments:
1. As a wrapped solution for the command pattern, the logger was leave out since must use the `Handler`'s class logger. Hence I just added the a second argument to pass the `TaggedLogger` object currently used for logging.
2. `hasActiveJoystickInput()` seems to be a good option.
3. True, I just added.
4. Complemented
5. I would maintain the intent of `registerScope()` to what it is. Instead, I added the `BootCommand` dispatch in `WenuLinkService` since its the responsible of triggering the boot and shutdown sequences.

Regards the ktlint format, I'm not used to create commits from any IDE, so I don't have the format triggering before commit. I run `ktlintFormat` manually, because there are some auto-fix that can be solve in better ways. I will pay more attention to it and prevent to ask for a review without formatting.

TimonSchreiber (approved):
Great work @angel-ayala, looks good to me now

PR #31, opened 2026-03-26
feature: fix busy-wait of requestShootPhoto

### LINKED ISSUE #30: CameraController: requestShootPhoto busy-waits between shots
In `CameraController.requestShootPhoto`, the interval between shots for multi-photo capture is implemented as a tight loop with no `delay()`, causing 100% thread utilization for the entire interval duration. Single-shot capture is unaffected.
Identified during review of #27 . Deferred from that PR as the fix required making `requestShootPhoto` suspendable and launching it via a coroutine scope, which was out of scope at the time.

Fix:
* Store the `CoroutineScope` as a field in `CameraHandler` and expose `launchCameraJob()`
* Make `requestShootPhoto` a suspend function with `delay(100)` in the wait branches
* Lunch it from `requestStartCapture` via `aircraft.cameras.launchCameraJob { ... }`

TimonSchreiber:
Fixes #30 

Addresses the busy-wait in requestShootPhoto identified during review of #27.

Changes:
- CameraHandler stores its CoroutineScope and exposes launchCameraJob()
- requestShootPhoto is now a suspend function with delay(100) in both wait branches
- requestStartCapture launches requestShootPhoto via launchCameraJob instead of calling it directly

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/camera/CameraHandler.kt
+++ b/app/src/main/java/org/WenuLink/adapters/camera/CameraHandler.kt
@@ -32,6 +32,8 @@ class CameraHandler {
     private var currentCommand: CameraCommand? = null
     private val commandChannel =
         Channel<Pair<CameraCommand, (String?) -> Unit>>(capacity = Channel.UNLIMITED)
+    private var handlerScope: CoroutineScope? = null
+    private var activeCameraJob: Job? = null
     var sequenceIndex = 0
     var captureTimestamp = System.currentTimeMillis()
     val lastCaptureMillis
@@ -163,5 +165,23 @@ class CameraHandler {
 
     fun canRecordVideo(cameraIdx: Int = 0): Boolean = CameraManager.canRecordVideo()
 
-    fun registerHandlerScope(scope: CoroutineScope) = startCommandProcessor(scope)
+    fun registerHandlerScope(scope: CoroutineScope) {
+        handlerScope = scope
+        startCommandProcessor(scope)
+    }
+
+    fun launchCameraJob(block: suspend () -> Unit): Job? {
+        activeCameraJob?.cancel()
+        return handlerScope
+            ?.launch { block() }
+            ?.also { job ->
+                activeCameraJob = job
+                job.invokeOnCompletion { activeCameraJob = null }
+            }
+    }
+
+    fun cancelCameraJob() {
+        activeCameraJob?.cancel()
+        activeCameraJob = null
+    }
 }
--- a/app/src/main/java/org/WenuLink/controllers/CameraController.kt
+++ b/app/src/main/java/org/WenuLink/controllers/CameraController.kt
@@ -15,6 +15,7 @@ import com.MAVLink.enums.STORAGE_TYPE
 import com.MAVLink.enums.STORAGE_USAGE_FLAG
 import io.getstream.log.taggedLogger
 import kotlin.math.roundToLong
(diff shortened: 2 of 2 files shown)


angel-ayala:
Not so sure if is a viable solution. If I'm not wrong, the MAVLink protocol requires to send `CAMERA_IMAGE_CAPTURED` every time that take a photo. Right now, the implemented logic is suitable for a single photo case, but definitively is not suitable for multiple photo.  I think that the most appropriate solution for this is to use the SDK's [`ShootPhotoAction`](https://developer.dji.com/api-reference/android-api/Components/Missions/DJIShootPhotoAction.html) since it allows both and possible can be reactive by using the listeners pattern avoiding while loops workload. I'm unsure if it could the definitive solution since is a `MissionAction` and I don't know if that is a common feature among the SDK compatible Aircrafts.

angel-ayala:
I was looking up at the SDK documentation and a proper mode must be set according to [ShootPhotoMode](https://developer.dji.com/api-reference/android-api/Components/Camera/DJICamera_DJICameraSettingsDef.html#djicamera_djicamerashootphotomode_inline). Additionally, there is a [SystemState](https://developer.dji.com/api-reference/android-api/Components/Camera/DJICamera.html#djicamera_setdjicameraupdatedsystemstatecallback_inline) that can be poll in the monitor loop to sync current camera state, possibly to trigger the required MAVLink message.

EDIT
------
I just found the [`FlatCameraMode`](https://developer.dji.com/api-reference/android-api/Components/Camera/DJICamera.html#djicamera_djiflatcameramode_inline) that can help to easily set camera modes. Additionally, the `PHOTO_INTERVAL`  mode states that:
> Sets the camera to take a picture (or multiple pictures) continuously at a set time interval. The minimum interval for JPEG format of any quality is 2s.

So even `delay(100)` patch suggested, can be easily increased to 1000.

TimonSchreiber:
I agree that is a good idea for handling multi-shot.
Since `ShootPhotoAction` is a a Mission Action, which may not be available on all compatible aircrafts, we might need a fallback anyway.
Do you want me to rework this pr to use the DJI SDK's interval shoot mode or handle this in a separate issue/pr?

angel-ayala:
True, we should need a fallback in case `MisionAction`s are not available.
Nevertheless, in both cases we must face the same issue, that is to capture the photoShoot event to send the `CameraImageCaptured` message and release the next photoShoot if needed as a event-driven approach. Event-driven here is key, because if use a plan flight where are `MisionAction`s under the hood, still need to capture such event and send the message. Additionally, I do not recommend to store CoroutineScope objects in handler os other classes because is prone to memory leak and you lost track of debugging when the scope is cancelled elsewhere. In contrast, Handler can deals with jobs or something similar.

Changes in #32 could alleviate this by leveraging the `WenuLinkHandler.monitorJob` loop to sync the camera state an update some `StateFlow` to call a previously defined callback from any controller, trigered each time an image is captured to further send the `CameraImageCaptured` through the controller.

In this regard, I think we should have two `CameraCommands` one to trigger a single photoShoot and wait for camera release to return and end the execution. The second class will deal with the required iterative call logic to shoot multiple photos, in this way the commands are triggering the events, and the WenuLinkHandler monitoring the state changes.

What do you think?

TimonSchreiber:
I agree with your suggested strategy of splitting the `CameraComand` into two separate commands to handle single and multi shot. Closing this pr in favour of a clean approach which utilizes the latest changes to `develop`.

PR #32, opened 2026-03-28
Refactor with command pattern in handlers, unify SDK entrypoint, and simplify initialization logics

angel-ayala:
Following changes aim to implement the general `CommandHandler` class to reuse the command pattern in `AircraftHandler`, `CameraHandler`, and `MissionHandler`, facilitating the request of asynchronous actions with the SDK. Now each `IHandler` that must deals with commands, must inherit from `CommandHandler`.

Current changes should allow to develop a more event-driven approach for #30 by leveraging [`MissionAction`](https://developer.dji.com/api-reference/android-api/Components/Missions/DJIMissionAction.html)s and state listeners.


angel-ayala:
Hi @TimonSchreiber , I just started by adding the command pattern in the handlers, and I saw better to include all major architecture changes, all reflected in the shared diagram.
In overall, now `AircraftHandler`, `MissionHandler`, and `CameraHandler` inherits from `CommandHandler`, to dispatch local commands specific to the handler. A new `WenuLinkHandler` was added as to dispatch specific commands to its respective handler, or execute cross-domain commands, inheriting from `RequestCommand`. e.g. to dispatch a `MissionCommand`, a proper dispatch of `StateTransition` on the `AircraftHandler`'s side must be done first before dispatch the `MissionHandler`'s command.

All MAVLink controllers that inherits from `IController`, now belongs to the `mavlink.controller` package. Since all controllers depends on `MAVLinkClient`, `mavlink` is the best place to have `controller` as sub package. Additionally, changes were made to enhance the modularization avoiding to call any class in the `parameters` package to enforce the `WenuLinkHandler` centralization. Now, `AircraftHandler` handles the parameters initialization avoiding the waiting mechanisms present in the controllers side.

`AircraftState` and `AircraftStateManager` now are more representative and handle the `ArduCopterFlightModes` and `ControlAuthorityTypes`, respectively. The main idea was to effectively comprise all information related to the current state that must be shared among the other classes, with updated information from the same place.
Similarly the new `APIManager` and `WenuLinkApp` classes,  were organized to enhance the SDK initialization  and expose useful information to the UI layer. `WenuLinkApp` now controls `WenuLinkService` and `WenuLinkHandler`, triggering proper init/shutdown sequences, and sharing information to the UI layer.

The current architecture is more robust and manageable in terms of validation and execution of different commands. The `WenuLinkHandler` now dispatch abstract `WenuLinkCommands` for all handlers that inherits from `CommandHandlers`, making the interaction of `MAVLinkController` and respective controllers through commands only. Such changes now reflect a cleaner distribution of concerns avoiding overloading certain classes.

Let me know if you can review it, thanks

TimonSchreiber (changes requested):
Nice work as always @angel-ayala , sorry for the delayed review.

The architectural changes move us into a great direction!
1 - 4 are points I think need to be fixed, the rest are more on the suggestion side:

### 1. `WenulinkHandler` never starts its own command processor

`WenuLinkHandler.registerScope` calls `startMonitorJob` but not `startCommandProcessor`:

```kotlin
override fun registerScope(scope: CoroutineScope) {
    aircraft.registerScope(scope)
    mission.registerScope(scope)
    camera.registerScope(scope)
    startMonitorJob(scope)      // no startCommandProcessor call
}
```

Because `WenuLinkHandler : CommandHandler<WenuLinkHandler>`, the base class owns a `commandChannel` that is only drained once `startCommandProcessor` is called. Without that call, any dispatch via the base-class `dispatchCommand(ICommand<WenuLinkHandler>, ...)` silently enqueues into a channel that is **never consumed**.

---

### 2. `isAircraftUp` returns a new `MutableStateFlow` instance on every call when the service is null

```kotlin
// ServicesViewModel
val isAircraftUp: StateFlow<Boolean>
    get() = (thisApp.wenuLinkService?.isAircraftBoot ?: MutableStateFlow(false))
```

This is a computed property, not a stable reference. When `wenuLinkService` is `null` (before start, or after stop), each invocation returns a **different** `MutableStateFlow(false)` object. Compose's `collectAsState()` captures the `StateFlow` reference at composition time. If the service is stopped and restarted, the composable continues collecting from the old (now-detached) `isAircraftBoot` on the destroyed service instance. The new service's flow is never observed, and the button state is permanently stale.

**Fix:** Expose a stable `MutableStateFlow<Boolean>` on `WenuLinkApp` (e.g., `val isAircraftBoot = MutableStateFlow(false)`), set it to `true` inside `WenuLinkService` after boot succeeds and to `false` inside `onDestroy()`, and have `ServicesViewModel` expose that stable application-level flow directly. This removes the dependency on a nullable service reference entirely.

---

### 3. Mission completion is not signaled, `controlAuthority` is permanently stuck in `WAYPOINT_MISSION` after a mission finishes

`MissionManager.addListeners` was simplified from three callbacks to one:

```kotlin
// Before
fun addListeners(onStart: () -> Unit, onWaypointReach: (Int) -> Unit, onFinish: (String?) -> Unit)

// After
fun addListeners(onWaypointStart: (Int) -> Unit)
```

The `onExecutionFinish` override in the new listener is a no-op. The old finish path called `controlTransition(ControlAuthority.NONE)` to reset authority after mission completion. That path no longer exists. After a waypoint mission finishes, `WenuLinkHandler.controlAuthority` remains `WAYPOINT_MISSION` indefinitely. Any subsequent `dispatchControlAuthority` call from a different mode sees `isNewAuthority == false` and returns early, blocking all further authority transitions until the service is restarted.

**Fix:** Re-introduce a finish callback, either restore an `onFinish` parameter to `MissionManager.addListeners`, or expose mission state as a `StateFlow` that `WenuLinkHandler` collects and reacts to with `dispatchControlAuthority(ControlAuthorityType.NONE)`.

---

### 4. `ActionCommand` registers callbacks it cannot deregister

`ActionCommand.execute()` registers two callbacks:

```kotlin
MissionActionManager.onFinish(action::class) {
    cont.resume(null)   // ActionCallbackKey returned but discarded
}
MissionActionManager.startListener {
    cont.resume("Action failed: $it")
}
```

`MissionActionManager.onFinish` now returns an `ActionCallbackKey` (good improvement), but `ActionCommand` discards it. The `invokeOnCancellation` handler calls `MissionActionManager.stop()` but does not remove the registered `onFinish` entry. On the next `ActionCommand` execution (e.g., a second `ReturnToHomeCommand`), the callback from the *previous* continuation is still registered. When the action finishes, it attempts to resume the already-completed continuation, throwing `IllegalStateException`.

**Fix:** Store the `ActionCallbackKey` returned by `onFinish` and call a `removeCallback(key)` inside both `invokeOnCancellation` and the `onFinish` handler itself (after resuming), to ensure one-shot semantics.

---

### 5. `permissionsList` is a public mutable list built in two stages

```kotlin
// WenuLinkApp: class-level initialization
val permissionsList = mutableListOf(/* base permissions */)
 
// WenuLinkApp.onCreate(): OS-version additions
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
    permissionsList += Manifest.permission.FOREGROUND_SERVICE
}
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
    permissionsList += Manifest.permission.POST_NOTIFICATIONS
}
```

Exposing `permissionsList` as a public `MutableList<String>` allows any caller to add or remove entries unintentionally.

**Suggestion:** Make `permissionsList` private, compute the full list eagerly in a single expression during `onCreate()` (or as a lazy property), and expose only `getMissingPermissions()` publicly.

---

### 6. Camera initialization is deferred out of the boot sequence, early camera queries will return `MAV_RESULT_UNSUPPORTED`

Previously `AircraftHandler.boot()` called `hasCameras = cameras.initCameras()` as part of the boot sequence. Now `CameraHandler.registerScope` launches it concurrently:

```kotlin
override fun registerScope(scope: CoroutineScope) {
    startCommandProcessor(scope, this@CameraHandler, logger)
    scope.launch { initCameras() }   // runs concurrently with boot
}
```

`WenuLinkHandler.registerScope` is called before `bootAircraft()`, so `initCameras()` races against the boot. If GCS sends `MAV_CMD_REQUEST_CAMERA_INFORMATION` during or immediately after boot, `handler.availableCameras` is likely still empty and `CameraController.reportCameras()` responds `MAV_RESULT_UNSUPPORTED`. GCS will think there is no camera.

**Suggestion:** Either `awaitCameraInit()` inside `WenuLinkHandler.bootAircraft()` before returning, or explicitly load cameras as a named step in the boot sequence with a logged result like the other boot steps.

---

### 7. `RequestTransition.execute()` wraps a synchronous state check in `suspendCancellableCoroutine` unnecessarily

```kotlin
override suspend fun execute(ctx: WenuLinkHandler): String? =
    suspendCancellableCoroutine { cont ->
        val prevState = ctx.aircraft.state.copy()
        ctx.aircraft.dispatchTransition(transition)
        val newState = ctx.aircraft.state.copy()
        cont.resume(if (prevState == newState) "State not changed" else null)
        cont.invokeOnCancellation { ctx.manualControl() }
    }
```

`dispatchTransition` is synchronous, the coroutine suspends and resumes inline without actually suspending. This is not wrong but misleads readers into thinking there is an async wait happening. A plain suspend function returning after the check would be clearer:
 
```kotlin
override suspend fun execute(ctx: WenuLinkHandler): String? {
    val prevState = ctx.aircraft.state.copy()
    ctx.aircraft.dispatchTransition(transition)
    return if (prevState == ctx.aircraft.state) "State not changed" else null
}
```
 
The cancellation hook (`ctx.manualControl()`) would need to be handled at the `RequestCommand` level instead (e.g., in `CommandHandler`'s cancellation path via `onStop()`).

---

### 8. `controllers.plusAssign(...)` is more verbose than `controllers +=` with no benefit

```kotlin
controllers.plusAssign(ConnectionController(client))
controllers.plusAssign(CommandController(client))
// ...
```

These are identical to `controllers += ConnectionController(client)`. The explicit form is harder to read and inconsistent with how the same pattern is used elsewhere in the codebase.

angel-ayala:
Great! I just addressed your comments:

1. Good catch! done
2. I didn't noticed, done
3. I leaved the single callbackEvent and improved the `WenuLinkHandler` to periodically sync the mission state and act accordingly at start, iteration, and finish the wp, an special sequence value = -1 indicates mission finish.
4. Just addressed, but is unfeasible to call `MissionActionManager.removeCallback(actionKey)` inside `onFinish` since is part of the declaration of `actionKey`.
5. True, the lazy property seems to be enough and 
6. I just added a flag from `CameraHandler` as part of `isAircraftPowerOn` to ensure boot of aircraft and camera before start any other service.
7. I totally agree, except the cancellation hook to be at `RequestCommand` level since is an abstract class.
8. I just discovered and found it fancy, but yes, is hard to see the assignation, I will be consistent with `+=`.

I still need to test it, but seems to be working fine.

app/src/main/java/org/WenuLink/sdk/APIManager.kt
```diff
@@ -0,0 +1,163 @@
...
+        logger.d { "Stopping.." }
+        tearDownListener()
+        DJISDKManager.getInstance().destroy()
+        sdkManagerCallback = null
+    }
+
+    fun registerCallbacks(
+        registrationCallback: (String?) -> Unit,
+        productConnectedCallback: (Boolean) -> Unit,
+        activationCallback: (Boolean, String?) -> Unit = { s, e -> },
```
TimonSchreiber:
```suggestion
        activationCallback: (Boolean, String?) -> Unit = { _, _ -> },
```

app/src/main/java/org/WenuLink/sdk/APIManager.kt
```diff
@@ -0,0 +1,163 @@
...
+        bindingCallback: (Boolean, String?) -> Unit = { s, e -> }
```
TimonSchreiber:
```suggestion
        bindingCallback: (Boolean, String?) -> Unit = { _, _ -> }
```

app/src/main/java/org/WenuLink/sdk/APIManager.kt
```diff
@@ -0,0 +1,163 @@
...
+    ) {
+        setActivationCallback(activationCallback)
+        setBindingCallback(bindingCallback)
+
+        sdkManagerCallback = object : DJISDKManager.SDKManagerCallback {
+            override fun onRegister(djiError: DJIError) {
+                if (djiError == DJISDKError.REGISTRATION_SUCCESS) {
+                    DJISDKManager.getInstance().startConnectionToProduct()
+                    loggerC.i { "Successfully registration" }
```
TimonSchreiber:
```suggestion
                    loggerC.i { "Successfully registered" }
```

TimonSchreiber (changes requested):
Well done @angel-ayala 

There is only one thing I would like you to take a second look at before merging.
The first time `onWaypointStart(-1)` is launched, both
`unvisitedSequence = true` **and** `mavlink = MISSION_STATE_COMPLETE`
are set. Now on the next loop of `monitorJob` the `missionHooks` function's `when` expression both cases are true, but only the first one gets executed, no call to `dispatchControlAuthority()`. Then `mission.stateSync()` is called but `fromMissionManager()` can only match the `else` case so we have both
`mavLink = MISSION_STATE.MISSION_STATE_UNKNOWN` **and** `unvisitedSequence = false`.
And we never get the chance to call `dispatchControlAuthority(ControlAuthorityType.NONE)` in `missionHooks`.

The easiest fix is probably to simply switch the order of the two cases in `missionHooks`:
```kotlin
private fun missionHooks() = when {
    mission.state.isComplete() -> dispatchControlAuthority(ControlAuthorityType.NONE)
    mission.state.unvisitedSequence ->
        logger.d { "Starting from new WP ${mission.state.sequence}" }
    else -> {}
}
```

One more thing for the future, it is better to `git rebase develop` onto the feature branch than using `merge`. Now the merge commit is part of this branch when it has absolutely nothing to do with this branch's development. You do have to add `--force-with-lease` the next time you push after `rebase`, but that shouldn't be a problem.

angel-ayala:
nice catch, definitively there was a miss implementation of the `unvisitedSequence` flag, also was odd to first launch hooks and after do the sync, common sense is to do the opposite, first sync and after trigger hooks. Now it should be working fine, when the `MissionManager` triggers a new WP number, `MissionHandler` will update it state with  `unvisitedSequence = true`  , `WenuLinkHandler` will sync the state and call `missionHooks()`, which now will call `processNode()` that do any require process, and set `unvisitedSequence = false` back.

This approach will be also useful for aircrafts non-compatible with [MissionControl](https://developer.dji.com/api-reference/android-api/Components/Missions/DJIMissionControl.html).

TimonSchreiber (approved):
Great work, ready for merge!

PR #37, opened 2026-04-03
Add missing message for test cases and other MissionAction commands

### LINKED ISSUE #19: Implement missing messages for Example Use Cases
The following messages are missing and needs to be implemented to comply with the [Example Use Cases](https://github.com/WenuLink/wenu-link-android/wiki/Example-Use-Cases):

- `NAV_DELAY`
- `CONDITION_YAW`
- `NAV_RETURN_TO_LAUNCH`

angel-ayala:
This PR aims to fix #19 by including the `NAV_DELAY`, `CONDITION_YAW`, and `NAV_RETURN_TO_LAUNCH` available through command and mission microservices.

Main changes were applied to the `mission` package by removing the old `MissionAction` action structure to reuse classes that inherits from `MissionActionCommand` to unify a modularized framework to launch SDK `MissionAction` through the command pattern. Each `MissionActionCommand` definition must interface the parameters from the MAVLink message through static method such as `fromCommandLong(...)` which must process the parameters and return the `MissionActionCommand` instance with proper arguments.


app/src/main/java/org/WenuLink/adapters/mission/MissionCommands.kt
```diff
@@ -158,14 +168,72 @@ data object ResumeActionCommand : MissionCommand {
...
+                timeMillis = delay
+            }
+            timeMillis *= 1000 // to ms
+            return DelayAction(timeMillis.toLong())
+        }
+
+        fun fromCommandLong(command: msg_command_long): DelayAction =
+            fromParameters(command.param1, command.param2, command.param3, command.param4)
+
+        fun fromMissionItem(missionItemMsg: msg_mission_item_int): MissionActionCommand =
```
TimonSchreiber:
```suggestion
        fun fromMissionItem(missionItemMsg: msg_mission_item_int): DelayAction =
```

app/src/main/java/org/WenuLink/adapters/mission/MissionCommands.kt
```diff
@@ -194,28 +262,123 @@ open class ActionCommand(val action: MissionAction) : MissionCommand {
...
+                )
+            val flightSpeed = if (commandIntMsg.param1 == -1f) 1f else commandIntMsg.param1
+
+            return RepositionAction(targetCoordinates, flightSpeed)
+        }
+
+        fun fromMissionItem(
+            missionItemMsg: msg_mission_item_int,
+            missionSpeed: Float
+        ): MissionActionCommand {
```
TimonSchreiber:
```suggestion
        ): RepositionAction {
```

app/src/main/java/org/WenuLink/adapters/mission/MissionCommands.kt
```diff
@@ -194,28 +262,123 @@ open class ActionCommand(val action: MissionAction) : MissionCommand {
...
+    private val isAbsolute: Boolean = true
+) : ActionCommand(
+    if (velocity == null) {
+        AircraftYawAction(angle, isAbsolute)
+    } else {
+        AircraftYawAction(angle, velocity)
+    }
+) {
+    companion object {
+        fun fromCommandLong(commandLongMsg: msg_command_long): MissionActionCommand {
```
TimonSchreiber:
```suggestion
        fun fromCommandLong(commandLongMsg: msg_command_long): RotateAction {
```

app/src/main/java/org/WenuLink/sdk/MissionManager.kt
```diff
@@ -140,24 +148,26 @@ object MissionManager {
...
+        is RotateAction ->
+            WaypointAction(WaypointActionType.ROTATE_AIRCRAFT, action.angle.roundToInt())
 
-        is MissionAction.GimbalPitch ->
-            WaypointAction(WaypointActionType.GIMBAL_PITCH, action.pitch)
+        is GimbalPitchAction ->
+            WaypointAction(WaypointActionType.GIMBAL_PITCH, action.angle.roundToInt())
 
-        MissionAction.TakePhoto ->
+        is ShootPhotoAction ->
```
TimonSchreiber:
```suggestion
        is PhotoAction ->
```

TimonSchreiber (changes requested):
In `MissionManager.mapAction` the case `is ShootPhotoAction` is unreachable since `ShootPhotoAction` does not implement `MissionActionCommand`. This should be `is PhotoAction` instead (see inline suggestion)

In the same function the `else` branch should probably log when an unknown action is encountered:

```kotlin
else -> {
    logger.w { "mapAction: unhandled action type ${action::class.simpleName}" }
    WaypointAction(WaypointActionType.STAY, 1000)
}
```

---

`DelayAction`, `RepositionAction`, and `RotateAction` all have one factory method that declares `MissionActionCommand` as their return type, but alway return a single concrete type, making the signature misleading by suggesting other types could be returned as well. (see inline suggestions)

---

*Suggestion:* Simplify `DelayAction.fromParameters`
The current implementation uses a variable named `timeMillis` but holds seconds until the very last line, which can be confusing. Something like this might be cleaner:

```kotlin
fun fromParameters(param1: Float, param2: Float, param3: Float, param4: Float): DelayAction {
    val delay = param1.toInt()
    val hours = param2.toInt()
    val minutes = param3.toInt()
    val seconds = param4.toInt()

    val totalSeconds = if (delay != -1) {
        delay
    } else {
        (if (hours != -1) hours * 3600 else 0) +
        (if (minutes != -1) minutes * 60 else 0) +
        (if (seconds != -1) seconds else 0)
    }
    return DelayAction(totalSeconds * 1000L)
}
```

The final `* 1000L` both converts to milli seconds and casts to `Long`.

---

*Note:* Both `CommandController.processLanding` and `.processReturn` now call `requestMode` directly instead of `dispatchCommand` like every other method there, is this intentional?

angel-ayala:
Nice catch @TimonSchreiber as always, indeed my bad with the return type and the wrong class.
And yes, `CommandController.processLanding` and `.processReturn` only request for a mode change, since `WenuLinkHandler` triggers both process in an asynchronous way as declared in `modeHooks()`. The implementation followed similar behavior that ardupilot firmware.
Now should be fine.

TimonSchreiber (approved):
Nice work @angel-ayala 
This looks good to me now

PR #38, opened 2026-04-07
feature: implement event-driven CAMERA_IMAGE_CAPTURED dispatch

### LINKED ISSUE #30: CameraController: requestShootPhoto busy-waits between shots

TimonSchreiber:
Closes #30

## Context

The original `requestShootPhoto` implementation drove photo capture from a busy-wait loop on the MAVLink processing thread, with no delay between iterations.
This was identified during review of #27 and deferred pending the architectural changes in #32. Now that `WenuLinkHandler.monitorJob` and `CommandHandler` are in place, a proper event-driven solution is possible.

### `CameraManager`
Registers a DJI `SystemState` listener on `init`. Detects the rising edge of `isStoringPhoto` (false -> true) and increments a `StateFlow<Int> captureCount` once per capture. Also adds `setIntervalSeconds`, `requestStartIntervalShoot`, and `requestStopIntervalShoot` to support interval shooting via the DJI SDK's native `INTERVAL` shoot mode.

### `CameraHandler`
Proxies `captureCount` from `CameraManager` and exposes `consumeCaptureEvent()`, which returns true once per new capture and increments `photoSeqIndex`.

### `WenuLinkHandler`
Adds a `cameraHooks()` branch to `startMonitorJob`, called in parallel with the existing aircraft and mission hooks. On each consumed capture event, invokes `onImageCaptured`, a nullable callback registered by `MAVLinkController` at startup.

### `CameraCommands`
Two new commands mirroring `StartRecordCommand`/`StopRecordCommand`:
- `StartIntervalShootCommand`: sets `ShootPhotoMode.INTERVAL`, configures the interval, and starts shooting.
- `StopIntervalShootCommand`: stops interval shooting and resets state.

### `CameraController`
`requestShootPhoto` and its busy-wait loop are removed. `requestStartCapture` now routes single-shot (`totalPhotos == 1`) to `TakePhotoCommand` and multi-shot to `StartIntervalShootCommand`.
`requestStopCapture` dispatches `StopIntervalShootCommand`. `CAMERA_IMAGE_CAPTURED` is no longer sent directly from `CameraController`,instead it is sent by `MAVLinkController` via the `onImageCaptured` hook for both single and interval shots.
mand and multi-shot to StartIntervalShootCommand

### DIFF
--- a/app/src/main/java/org/WenuLink/sdk/CameraManager.kt
+++ b/app/src/main/java/org/WenuLink/sdk/CameraManager.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.graphics.SurfaceTexture
 import android.media.MediaFormat
 import dji.common.camera.SettingsDefinitions
+import dji.common.camera.SystemState
 import dji.common.error.DJIError
 import dji.common.util.CommonCallbacks
 import dji.sdk.camera.Camera
@@ -12,9 +13,6 @@ import dji.sdk.codec.DJICodecManager
 import io.getstream.log.taggedLogger
 import java.nio.ByteBuffer
 import kotlin.coroutines.resume
-import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.flow.StateFlow
-import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.suspendCancellableCoroutine
 import org.WenuLink.adapters.AsyncUtils
 
@@ -40,9 +38,6 @@ object CameraManager {
     var fwVersion: String? = null
     var cameraMode = SettingsDefinitions.CameraMode.UNKNOWN
     var captureMode = SettingsDefinitions.ShootPhotoMode.UNKNOWN
-    private val _captureCount = MutableStateFlow(0)
-    val captureCount: StateFlow<Int> = _captureCount.asStateFlow()
-    private var wasStoringPhoto = false
 
     @Synchronized
     fun init(camera: Camera) {
@@ -56,20 +51,14 @@ object CameraManager {
         frameWidth = resolution[0].toInt()
         frameHeight = resolution[1].toInt()
 
-        frameRate = camera.capabilities.videoResolutionAndFrameRateRange()[0].frameRate.toString()
+        frameRate = camera.capabilities.videoResolutionAndFrameRateRange()[0].frameRate
+            .toString()
             .replace("FRAME_RATE_", "")
             .replace("_FPS", "")
             .replace("_DOT_", ".")
             .toFloat()
 
         logger.i { toString() }
-
-        camera.setSystemStateCallback { state ->
-            if (state.isStoringPhoto && !wasStoringPhoto) {
-                _captureCount.value += 1
-            }
-            wasStoringPhoto = state.isStoringPhoto
-        }
     }
 
     fun createCompletionCallback(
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/WenuLink/sdk/CameraManager.kt
```diff
@@ -319,11 +326,7 @@ object CameraManager {
         return suspendCancellableCoroutine { cont ->
             camera.stopRecordVideo(
                 SDKUtils.createCompletionCallback { error ->
```
angel-ayala:
```suggestion
                SDKUtils.createCompletionCallback(cont::resume)
```

angel-ayala (changes requested):
Nice done @TimonSchreiber , this was exactly what I thought. The implementation is clean an easy to follow.
My only observation is related to the event-driven approach, and the unexpected interaction with `CameraManager` by using `StateFlow` to update the variable from the sdk execution thread.
Therefore I can comment out two things, the first related to the interaction with the Manager classes since, all previous cases define a `registerStateCallback` method to interface the SDK method with extra logic if required.  In this regard, all SDK related Manager keep as clean as possible exposing fixed entry points to interact with the SDK, being as "dumb" as possible. Hence, all required logic must be defined in the proper handlers.

Second, if moving the state callback registration logic to `CameraHandler` the `captureCount` definition as `StateFlow` can be directly exposed to `WenuLinkHandler`, using it in a more higher level on the architecture, considering a bottom-up layers definition as SDK -> Handlers -> Controllers/Services -> Application/UI. In this regard, the `registerCameraHooks` definition can be moved inside `CameraController` where it belongs. `CameraController` already has a similar purpose argument named `onSetMessageRate` and possibly can be similarly set or by leveraging `captureCount` asynchronous triggering.

Everything else looks good

TimonSchreiber:
Good call @angel-ayala !

I couldn't find a workaround to passing `WenuLinkHandler` as a temporay field to `CameraController` for initializing the `onCameraCapture` function. But since every `IController` method uses the `WenuLinkHandler` anyway, maybe it should also become a permanent field like `client`:

```kotlin
interface IController {
    val client: MAVLinkClient
    val handler: WenuLinkHandler

    fun processMessage(msg: MAVLinkMessage): Boolean =
        // TODO: centralize client sending answer messages
        false

    // and so on...
```

If we do this, I would not add those changes to this pr but open a new one. What do you think?

angel-ayala (changes requested):
That refactoring seems to be a good enhancement, but I don't see it necessary to solve the current issue. Actually, I see that the problem right now, is not having or not `WenuLinkHandler` as a `IController` attribute, but instead, where must be set the `onImageCaptured` callback.

I think is more efficient to control a register and "unregister" such callback to have it active only when necessary. That means that a good approach could be to set `onImageCaptured` inside `requestStartCapture`, and unset inside `requestStopCapture`. For cases of single capture, inside the `dispatchCommand` callback in  `requestStartCapture` must be enough:

```kotlin
handler.dispatchCommand(command) { error ->
    if (error != null) logger.w { "requestStartCapture error: $error" }
    onImageCaptured = null
}
```

Moving `WenuLinkHandler` inside `IController` would be a good enhancement to implement in the near future. We should discuss priorities to close a v1 and leave everything is not a priority for next versions.

TimonSchreiber:
Great idea @angel-ayala , thats a lot simpler than what I had come up with!
I agree that we do not have to concern us with `WenuLinkHandler` being an attibute of `IController` at the moment, especially since your solution removed the immediate need for it, but in the future I think it is the right thing to do.

angel-ayala (approved):
Yes @TimonSchreiber , I totally agree that is necessary to refactor `IController` , but in another PR, this one seems good to me, can you open another issue and mention it please? thanks.

PR #39, opened 2026-04-07
Refactor/numeric type casts

TimonSchreiber:
I noticed some redundant type conversions during my last review.
Kotlin allows implicit widening through operator overloading `Int / Float = Float`, I hope this makes some expressions easier to read.
No changes in behaviour


angel-ayala (approved):
makes sense to me, nice done.

PR #41, opened 2026-04-07
feature: fix takeoff not using the correct altitude

### LINKED ISSUE #36: No parameters reading for MAV_CMD_NAV_TAKEOFF
When dealing with the `MAV_CMD_NAV_TAKEOFF` no messages parameters are not being read. Hence, desired take off altitude is not used, reaching always the same altitude when take off.

TimonSchreiber:
Closes #36 

### `CommandController`:
Passes `commandMsg.param7` (altitude) to `TakeoffCommand`. Other values might follow later

### `AircraftCommand`:
Calls `AircraftHandler.goToAltitude(initialAltitude)` after successful takeoff

### `AircraftHandler`:
Adds a `goToAltitude(altitude, speed)` method called after `takeOff()` and `isFlying` is established.

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/RequestCommand.kt
+++ b/app/src/main/java/org/WenuLink/adapters/RequestCommand.kt
@@ -90,12 +90,13 @@ data class RequestTakeoff(
                     cont.resume(error)
                     return@dispatchCommand
                 }
+
+                val coordinates = ctx.aircraft.currentCoordinates
+                    ?: return@dispatchCommand cont.resume("No aircraft position available")
+
                 ctx.dispatchCommand(
                     WenuLinkCommand.Mission(
-                        RepositionAction(
-                            ctx.aircraft.getCurrentCoordinates()!!.copy(alt = altitude),
-                            speed
-                        )
+                        RepositionAction(coordinates.copy(alt = altitude), speed)
                     ),
                     cont::resume
                 )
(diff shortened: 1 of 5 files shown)


angel-ayala (changes requested):
Seems better to implement a `RequestCommand` that could triggers the `aircraft.TakeoffCommand`, followed by `mission.RepositionAction`.

Maybe we should define that only `RequestCommand` can be called from outside the `adapters` package.

TimonSchreiber:
Thats probably a cleaner solution than the one I proposed initially. Please take a look if this is what you had in mind.

I couldn't find a way to limit the access of the Command classes to the `adapter` package, the keyword `internal` limits the access to the own module, but since the wole application is one module, this wouldn't help

angel-ayala (changes requested):
That was exactly what I meant, the implementation is correct with two minor comments.


1. In `AircraftHandler` you have unnecessary imports.
2. In the `RequestTakeoff`, you are using a not-null assertion `!!` and must be avoided.
```
RepositionAction(
    ctx.aircraft.getCurrentCoordinates()!!.copy(alt = altitude),
    speed
)
```
I remember that `AircraftHandler.getCurrentCoordinates` method was implemented long time ago and now seems a good time to be updated. Possibly, instead we can have a property value `currentCoordinates` (just an idea), with the `get()` method to read such information from telemetry.

TimonSchreiber:
Good catch on the imports, I though I had set up Android Studio to optimize imports automatically.
Replacing `getCurrentCoordinates` with a class attribute is a good idea and was definitely overdue since `FCManager` was still used for it.

angel-ayala (approved):
All set, good job.

PR #42, opened 2026-04-07
Fix: Replace unbounded waitReady with waitTimeout in AircraftHandler

TimonSchreiber:
`awaitFlightState`, `loadParameters`, and `waitAndConfirmLanding` previously used `AsyncUtils.waitReady`, which blocks indefinitely if the expected condition is never met. A stuck takeoff, failed parameter load, or missed landing confirmation threshold would hang the respective command permanently with no way to recover.
All three now use `waitTimeout` with explicit timeouts:

- `awaitFlightState`: 15s default, propagated from `TakeoffCommand`
- `loadParameters`: timeout propagated from `boot()`, consistent with other boot steps
- `waitAndConfirmLanding`: 30s default; on timeout, `confirmLanding` is still attempted since the drone is likely near the ground

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
+++ b/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
@@ -275,19 +275,18 @@ class AircraftHandler : CommandHandler<AircraftHandler>() {
         return flyingStateUpdated
     }
 
-    suspend fun waitAndConfirmLanding(timeout: Long = 30_000L) {
+    suspend fun waitAndConfirmLanding() {
         // https://developer.dji.com/api-reference/android-api/Components/FlightController/DJIFlightController.html#djiflightcontroller_confirmlanding_inline
         logger.d { "\tWaiting altitude of 0.3m" }
-        val confirmationNeeded = AsyncUtils.waitTimeout(
-            timeout,
-            100L,
-            FCManager::needLandingConfirmation
-        )
+        AsyncUtils.waitReady(100L) {
+            FCManager.needLandingConfirmation() ||
+                FCManager.getAltitude() < 0.5f
+        }
 
-        if (!confirmationNeeded) logger.w { "Landing confirmation timeout, attempting anyway" }
-
-        FCManager.confirmLanding {
-            logger.d { "\tLanding confirm" }
+        if (FCManager.needLandingConfirmation()) {
+            FCManager.confirmLanding { logger.d { "\tLanding confirmed" } }
+        } else {
+            logger.w { "Landing confirmation skipped: aircraft already on ground" }
         }
     }
 }
--- a/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
+++ b/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
@@ -262,20 +262,20 @@ class AircraftHandler : CommandHandler<AircraftHandler>() {
         FCManager.startTakeoff()
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
```diff
@@ -272,26 +269,32 @@ class AircraftHandler : CommandHandler<AircraftHandler>() {
...
         // https://developer.dji.com/api-reference/android-api/Components/FlightController/DJIFlightController.html#djiflightcontroller_confirmlanding_inline
         logger.d { "\tWaiting altitude of 0.3m" }
-        AsyncUtils.waitReady(100L, FCManager::needLandingConfirmation)
+        val confirmationNeeded = AsyncUtils.waitTimeout(
+            timeout,
+            100L,
+            FCManager::needLandingConfirmation
+        )
+
+        if (!confirmationNeeded) logger.w { "Landing confirmation timeout, attempting anyway" }
```
angel-ayala: here I think should be kept the `waitReady` since is an altitude related waiting mechanism, from my understanding. The `FCManager` will only set `needLandingConfirmation == true` when it actually reaches the minimum altitude before asking user's confirmation, that is being automated here.
Therefore, any call of `FCManager.confirmLanding` before `needLandingConfirmation == true` is invalid.

If any errors happens and `needLandingConfirmation` never became `true`, is better to wait for a low altitude, before `waitTimeout` of `needLandingConfirmation`, to finish the execution.

Thinking in other possibilities, maybe an asynchronous not-blocking mechanism can be implement on the `WenuLinkHandler.monitorJob` by iteratively updating the `needLandingConfirmation` state, and trigger a corresponding hook.

angel-ayala (changes requested):
This is really a good move! However, I would like to ask you to review the changes in `waitAndConfirmLanding` since the premature call of `FCManager.confirmLanding` is invalid. I think that a better approach for this waiting mechanism is to performs periodical updates and act accordingly on `WenuLinkHandler`. I pointed out a suggestion with more details.

TimonSchreiber:
Your objection is correct, `FCManager.confirmLanding` should only be called when `FCManager.needLandingConfirmation() == true`. I added an alternative exit condition comparing the current altitude in case an error causes `isLandingConfirmationNeeded` to never be set to `true`

angel-ayala (changes requested):
Those changes still not solve the possible premature call of `FCManager.confirmLanding()` since `FCManager.needLandingConfirmation()` is very likely to became `true` with ~0.3m altitude. I suggest that instead one waiting line, split it in two, one unbounded `waitReady` for minimal altitude, and another bounded `waitTimeout` for the `needLandingConfirmation`:

```kotlin
AsyncUtils.waitReady(100L) { FCManager.getAltitude() < 0.5f }
val confirmationNeeded = AsyncUtils.waitTimeout(100L, 5000, FCManager::needLandingConfirmation)

if (!confirmationNeeded) logger.w { "Landing confirmation timeout, attempting anyway" }
else FCManager.confirmLanding {
    logger.d { "\tLanding confirmed" }
}
```

I such way reaching a low altitude and avoid unbounded wait logic. If I'm not wrong, the delay function inside `AsyncUtils.waitReady` and `AsyncUtils.waitTimeout` are already suspendable, so if cancel the `LandCommand`, which do not exists yet, must properly handle what happens next.

angel-ayala (dismissed):
Well done!

TimonSchreiber:
Had to rebase this branch, please approve again @angel-ayala

angel-ayala (approved):
Sure

PR #43, opened 2026-04-09
Add centralized MAVLink message classes

TimonSchreiber:
## Summary

MAVLink command handlers throughout the codebase extracted positional parameters inline using raw param1…param4 accesses with no indication of what each position means. Understanding any handler required cross-referencing the MAVLink spec externally.

This PR introduces a mavlink/params/ package with one data class per MAV_CMD that the application handles. Each class:
- names every field after the MAVLink spec parameter description
- carries a KDoc link to the relevant spec entry
- provides a from() factory for each applicable message type (msg_command_long, msg_command_int, or msg_mission_item_int)

## Notable changes

| | |
| --- | --- |
| `ConditionYawParams` | Fixes inverted `relative` boolean, original `param4.toInt() == 0` treated `0` as relative, which is backwards per the spec (`MAV_BOOL_FALSE = 0 = absolute`) |
| `MissionHandler` | `assembleTakeoffNode()` helper extracted for consistency with `assembleWaypointNode()`; `MessageUtils` import eliminated |
| `RepositionAction` | `fromMissionItem()` removed, had no call site and is superseded by `NavWaypointParams` |
| `DelayAction` / `RotateAction` | `fromCommandLong()` / `fromMissionItem()` factories replaced by `fromParameters(NavDelayParams)` / `fromParameters(ConditionYawParams)` |

## Files changed

| File | Change |
| --- | --- |
| `mavlink/params/CameraParams.kt` | New - 6 commands |
| `mavlink/params/FlightParams.kt` | New - 2 commands |
| `mavlink/params/NavigationParams.kt` | New - 7 commands |
| `mavlink/params/SetMessageIntervalParams.kt` | New - 1 command |
| `mavlink/params/ParamUtils.kt` | New - shared MAV_BOOL helper |
| `mavlink/controllers/CameraController.kt` | Migrated to params classes |
| `mavlink/controllers/CommandController.kt` | Migrated to params classes |
| `mavlink/controllers/NavigationController.kt` | Migrated to params classes |
| `mavlink/controllers/TelemetryController.kt` | Migrated to params classes |
| `adapters/mission/MissionHandler.kt` | Migrated to params classes |
| `adapters/mission/MissionCommands.kt` | Migrated to params classes |

## Testing

No behavioral changes intended except the `ConditionYAW` boolean fix noted above.

## Note

`SetMessageIntervalParams.kt` was meant to be named `TelemetryParams.kt` but linter insists that in a file with only one class, the file and class have the same name.


angel-ayala (commented):
Hi @TimonSchreiber , I think that understand the motivation behind this changes, and makes sense. However, the implementation appears to be an extra layer to do almost the same, to associate the command parameters value from the MAVLink message structure into a meaningful variable.
Nevertheless, you are kind of duplicating the messages with the static `from(...)` method that reads the message structure and associate required parameters as attributes, for further use with a proper variable name and easy to read. For me this seems to be a class redefinition with a distant message relation.

Additionally, looking the overall application, we actually must address something in this regard, but needs to be in both ways, for reading an interpret incoming values and for creating and sending message. Of course, not all messages are being sent or read, but the same logic applies. A message parameters translation is required to acquire arguments, or prepare telemetry values into messages object.

Probably we should create a messages registry, or some factory patter,with a base `WenuLinkMessage` class for example, to abstract all the required message processing as possible, in addition to inherit each related MAVLink message class, wrapping the proper logic in a self-contained manner.

TimonSchreiber:
I see your point @angel-ayala , only centralizing the decoding of mavlink messages doesn't solve the problem. Would you consider expanding these classes with a `toMavLink()` method a suitable solutuion or do you want to go a different direction? Something like this:

```kotlin
class WenuLinkImageStartCapture(
    val targetCameraId: Int,
    val intervalSec: Float,
    val totalImages: Int,
    val sequenceNumber: Int,
) : WenuLinkMessage {
    constructor(msg: msg_command_long) : this(
        targetCameraId = msg.param1.toInt(),
        ...
    )

    fun toMavLink(): MAVLinkMessage= msg_command_long().apply {
        command = MAV_CMD.MAV_CMD_IMAGE_START_CAPTURE
        param1 = targetCameraId.toFloat()
        ...
    }
}
```

---

Either way, I did find a genuin bug that this pr also solves: the `MissionCommands.RotateAction` incorrectly decodes `param4 = 0` to mean relative when the documentation says that a value of 0 mean absolute and 1 means relative.

```kotlin
val relative = commandLongMsg.param4.toInt() == 0 // <-- param4 == 0 means relative is true
 // convert to compatible intervals
angle = ((angle + 180) % 360) - 180
if (!relative) return RotateAction(angle, null, true)

// clip angular speed
angularSpeed = angularSpeed.coerceIn(0f..100f)
return RotateAction(angle, angularSpeed, false)
```

I can either create a separate pr just to fix this bug or we could accept this pr for now and expand to a full registry later.

angel-ayala:
Hi @TimonSchreiber , sorry for the delay response, but yes, definitively something like that. However, I think that will require different abstraction strategies to avoid the definition of all messages. Maybe some dataclass structure that receives the message ID, and a pair of key-value to define a more meaningful translation of corresponding param1..7 of each `COMMAND_LONG`. I think that you could start with the command type messages before the telemetry related ones to stabilize the architecture first. I like the idea of having `WenuLinkMessage` as base class, and `WenuLinkImageStartCapture` can be something with a stronger association with MAVLink such as `ImageStartCaptureCommandLong`.

Then an auxiliary function could do the mapping as required, as you did. I just remember that in some moment I mentioned something to have a centralized message processing in the sending operation to ensure a more robust routing strategy. Maybe this changes could be organized to address a message handling logic considering the case discussed until now, and the addition of an improved routing system since all is related. However, changes for the routing systems will be applied in other PR.

What do you think?

EDIT
---------
I was having a second thought and possibly the controllers classes must implement their own message registry. Right now all classes that inherits from `IController` act as a distributed message processing logic, with a specific message set available from a single controller. Such registry must define different message type for the `processMessage`, `processCommandLong`, `processCommandInt`, `processRequestLong`, `processRequestInt`, and `createMessage` methods. Possibly, this changes will require a new `messages` package inside `mavlink` instead the `params` as proposed.

angel-ayala:
Discussed changes are definitively required to parse message into something meaningful for the command architecture. Last test of #50 are returning `Cannot round NaN value` in two different situations, the first one was encountered in `NavigationController.msgRawGPSInt()` when created which trigger an exception on the `MAVLinkClient` side, and the second one during flight plan upload, crashing the application.
Here are the logs line for the last case:

```plain
2026-04-12 23:24:02.853 14365-14757 NavigationController    org.WenuLink                         D  (DefaultDispatcher-worker-8:344) processMissionItem
2026-04-12 23:24:02.853 14365-14757 NavigationController    org.WenuLink                         D  (DefaultDispatcher-worker-8:344) 	MAVLINK_MSG_ID_MISSION_ITEM_INT - sysid:255 compid:190 param1:0.0 param2:0.0 param3:0.0 param4:0.0 x:0 y:0 z:0.0 seq:6 command:20 target_system:1 target_component:1 frame:2 current:0 autocontinue:1 mission_type:0
2026-04-12 23:24:02.853 14365-14757 MissionHandler          org.WenuLink                         D  (DefaultDispatcher-worker-8:344) Append mission item.
2026-04-12 23:24:02.854 14365-14757 NavigationController    org.WenuLink                         D  (DefaultDispatcher-worker-8:344) ackReceivedItem #6
2026-04-12 23:24:02.856 14365-14365 MissionHandler          org.WenuLink                         D  (main:2) Executing: UploadMissionCommand
2026-04-12 23:24:02.864 14365-14365 MissionHandler          org.WenuLink                         D  (main:2) Command failed: UploadMissionCommand
2026-04-12 23:24:02.864 14365-14365 NavigationController    org.WenuLink                         D  (main:2) sendStatusText
2026-04-12 23:24:03.095 14365-14623 DJIUSBWifiSwitchManager org.WenuLink                         E  ****resetSwitchFromWifiFlag
2026-04-12 23:24:03.098 14365-14623 DJIUsbAccessoryReceiver org.WenuLink                         E  receiver destroySession
--------- beginning of crash
2026-04-12 23:24:03.196 14365-14365 AndroidRuntime          org.WenuLink                         E  FATAL EXCEPTION: main
                                                                                                    Process: org.WenuLink, PID: 14365
                                                                                                    java.lang.IllegalArgumentException: Cannot round NaN value.
                                                                                                    	at kotlin.math.MathKt__MathJVMKt.roundToInt(MathJVM.kt:1165)
                                                                                                    	at org.WenuLink.sdk.MissionManager.mapAction(MissionManager.kt:156)
                                                                                                    	at org.WenuLink.sdk.MissionManager.uploadMission(MissionManager.kt:135)
                                                                                                    	at org.WenuLink.adapters.mission.UploadMissionCommand.execute(MissionCommands.kt:42)
                                                                                                    	at org.WenuLink.adapters.mission.UploadMissionCommand.execute(MissionCommands.kt:31)
                                                                                                    	at org.WenuLink.commands.CommandHandler$startCommandProcessor$1$deferred$1.invokeSuspend(CommandHandler.kt:36)
                                                                                                    	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                                                                                                    	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
                                                                                                    	at android.os.Handler.handleCallback(Handler.java:942)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:99)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:211)
                                                                                                    	at android.os.Looper.loop(Looper.java:300)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:8232)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:580)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1028)
```

probably main cause is by the availability of latitude and longitude coordinates and GPS fix level in both situations, by sending coordinates in `GPS_RAW_INT` message and by receiving coordinates inside an `MISSION_ITEM_INT` message.

Such error makes sense to me, because on the first case in `NavigationController.msgRawGPSInt()` there is no coordinates to send, if GPS has not acquired a position yet. On the second case, definitively should be something like a misaligned or outdated processing form.

angel-ayala (changes requested):

TimonSchreiber:
Sorry for the late follow up, I wasn't sure how to go forward.
I hope this is what you had in mind @angel-ayala : for now each controller holds their own registries, but in the future it could possibly move to `MavLinkController`

app/src/main/java/org/WenuLink/mavlink/messages/MessageUtils.kt
angel-ayala, line 1: There is another `MessageUtils` in `adapters.WenuLinkUtils.kt` maybe should be moved in here.

angel-ayala (changes requested):
Is very close to what I had in mind indeed, I was actually thinking to extend the MAVLinkMessage class, but now I'm not so sure how that could be addressed. Actually, now I was thinking in a major refactor that could possibly enhance this aspects, for example, similar has we did with the commands, we could have a self-contained message processing logic for both incoming and outgoing direction. In such way controllers will work as message providers to enable/disable certain messages types. I think this way message processing logic is all defined in the same place, instead of being in `controllers` and `messages` as is right now.

I you agree @TimonSchreiber with such major refactor, I can open another issue to discuss how to address the changes in a future.
Right now we can work on top of this structure you has been developed, since its already present a robust definition for command messages parameters. The only thing, is that we should be consistent in having a controller and message pair of classes.
So, for the case of `TelemetryMessages` maybe you can complement it with the class for the `msg_request_data_stream` message by addressing the `timeInterval` logic inside the same file to be able to rename it.

Another change that I see necessary, is to move the `MessageUtils` object from `WenuLinkUtils.kt` to the new `messages` package.
Everything else looks good to me.

TimonSchreiber:
I think any more changes will go over the scope of one pr, so opening an issue for discussion is a good idea.
Good idea extracting `msg_request_data_stream` so the class names can all match their counterpart and good catch with `MessageUtils` in `WenuLinkUtils.kt`

angel-ayala (dismissed):
Nice! I will open the issue then. Thanks!

TimonSchreiber:
Just rebased this feature branch

angel-ayala (dismissed):
All good

TimonSchreiber:
Had to resolve merge conflicts again...

angel-ayala (approved):

angel-ayala:
Oh now I understood, yep, maybe we could start to prioritize PR merges.

PR #45, opened 2026-04-09
feature: add WenuLinkHandler to IController

### LINKED ISSUE #44: Refactor IController to have WenuLinkHandler as an attribute
Many if not most methods defined in the classes that implement `IController` use an instance of `WenuLinkHandler`. Instead of passing it around every time a Controller method is called, do it once at construction.

TimonSchreiber:
Closes #44 

- Add `WenuLinkHandler` to `IController`

- Update `MAVLinkController`

- Adjust all Controller classes


angel-ayala (approved):
Good work!

PR #48, opened 2026-04-10
Replace String? success/failure pattern with CommandResult

### LINKED ISSUE #40: ICommand Return Type
The current pattern of using `string?` in the return type for methods like `ICommand.validate()` and `execute()`, `CommandHandler.stopCommand()` and `dispatchCommand`, and their related implementations harbours two problems.
First of all it is a known anti-pattern where `null` means success and any other `string` is an error, this would be surprising to any developer seeing it for the first time. And while currently not in use, it prevents us from ever returning meaningful data for the success path, which might be useful in the future.

I would like to propose to use a `sealed class` as the return type with explicit `Success` and `Failure` variants which would make the code easier to read and gives us the most flexibility going forward.
Something like `CommandResult` for example:

```kotlin
sealed class CommandResult<out T> {
    data class Success<T>(val value: T) : CommandResult<T>()
    data class Failure(val reason: String) : CommandResult<Nothing>()
}
```

Since this is a broader refactor and touches central elements like `ICommand`, I wanted to discuss this with you @angel-ayala first before jumping into action. With the architecture now stable post #37, this is the right moment. The necessary workload to implement this will only grow as more commands are added.

TimonSchreiber:
Closes #40 

Introduce a sealed class CommandResult<T> with Success and Failure variants, replacing the anti-pattern of null meaning success and a non-null String meaning failure. Also adds a UnitResult typealias for the common CommandResult<Unit> case.

Affected: ICommand, CommandHandler, all command implementations (AircraftCommands, MissionCommands, CameraCommands), AircraftState transitions, AircraftHandler, WenuLinkHandler, RequestCommand, and all MAVLink controller call sites.

The sdk layer (CameraManager, FCManager, MissionManager) intentionally retains String? as it reflects the DJI SDK error type directly.

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/RequestCommand.kt
+++ b/app/src/main/java/org/WenuLink/adapters/RequestCommand.kt
@@ -18,7 +18,9 @@ import org.WenuLink.adapters.mission.MissionCommand
 import org.WenuLink.adapters.mission.RepositionAction
 import org.WenuLink.adapters.mission.ReturnAction
 import org.WenuLink.adapters.mission.StartWaypointMission
+import org.WenuLink.commands.CommandResult
 import org.WenuLink.commands.ICommand
+import org.WenuLink.commands.UnitResult
 
 sealed interface WenuLinkCommand {
     data class Aircraft(val command: AircraftCommand) : WenuLinkCommand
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/WenuLink/adapters/RequestCommand.kt
```diff
@@ -28,38 +30,40 @@ sealed interface WenuLinkCommand {
...
 }
 
 data class RequestLand(val withLandingConfirmation: Boolean = true) :
     RequestTransition(LandTransition) {
-    override suspend fun execute(ctx: WenuLinkHandler): String? {
-        val transitionError = super.execute(ctx)
-        if (transitionError != null) return transitionError
+    override suspend fun execute(ctx: WenuLinkHandler): UnitResult {
+        val transitionResult = super.execute(ctx)
+        if (transitionResult is CommandResult.Failure) return transitionResult
```
angel-ayala: can be this way instead?

```suggestion
        if (transitionResult.hasError) return transitionResult
```

app/src/main/java/org/WenuLink/adapters/RequestCommand.kt
```diff
@@ -28,38 +30,40 @@ sealed interface WenuLinkCommand {
...
 
         ctx.dispatchControlAuthority(ControlAuthorityType.TIMELINE_COMMAND)
 
         return suspendCancellableCoroutine { cont ->
-            ctx.dispatchCommand(WenuLinkCommand.Mission(LandAction(true))) { error ->
-                if (error != null) {
-                    cont.resume(error)
+            ctx.dispatchCommand(WenuLinkCommand.Mission(LandAction(true))) { result ->
+                if (result is CommandResult.Failure) {
+                    cont.resume(result)
```
angel-ayala:
```suggestion
                if (result.hasError) {
                    cont.resume(result.reason)
```

app/src/main/java/org/WenuLink/adapters/WenuLinkHandler.kt
```diff
@@ -155,13 +157,13 @@ class WenuLinkHandler : CommandHandler<WenuLinkHandler>() {
             }
         }
 
-    suspend fun bootAircraft(): String? {
-        val bootError = dispatchAndAwait(WenuLinkCommand.Aircraft(BootCommand(10000L)))
-        if (bootError == null) {
+    suspend fun bootAircraft(): UnitResult {
+        val bootResult = dispatchAndAwait(WenuLinkCommand.Aircraft(BootCommand(10000L)))
+        if (bootResult is CommandResult.Success) {
```
angel-ayala:
```suggestion
        if (bootResult.isOk) {
```

app/src/main/java/org/WenuLink/commands/CommandResult.kt
angel-ayala, line 1: I think this one must be updated to fit the suggestions.

angel-ayala (commented):
Well done @TimonSchreiber ! changes seems good to me, and I would like to suggest to update how check the outcome of the new `CommandResult`. I saw that you are currently comparing if the result object is the `Failure` or the `Success` class, but is odd to me and unnecessary, I think that is more adequate to have simpler comparison such as `CommandResult.isOk` or `CommandResult.hasError`. Possibly with a errorReason attribute, that is being set only with Failure init, can be used to define both useful flags.

In the suggestions you can find more concrete examples.

TimonSchreiber:
Good idea @angel-ayala this makes it a lot easier to read. I only left the smart cast `result is CommandResult.Failure` in the `NavigationController.processMissionItem` method, since we need it there for the `sendStatusText` call:

```kotlin
if (result is CommandResult.Failure) {
    sendStatusText(
        result.reason,
        MAV_SEVERITY.MAV_SEVERITY_ERROR
    )
}
```

TimonSchreiber:
It seems that I made a mistake when rebasing this branch and now all changes from #49 are also in here, sorry for that.

angel-ayala:
Cool, definitively now seems easier to follow. Nevertheless, I think that such case you mention of keeping `result is CommandResult.Failure` runs out of the architecture scope, a better approach to address this I think can be in the `CommandResult` object with this update:
```kotlin
val errorReason: String
    get() = (this as? Failure)?.reason ?: "Unknown reason"
```
With this I think can be avoided the smart cast case on the comparison, wrapping everything up with the underlying logic.

PS: No worries about the changes duplication, only ensure that nothing was leaved out.

TimonSchreiber:
Done, but I think the fallback value should show that it can only be reached in case `errorReason` was called on a `CommandResult.Success`, so I went with `"errorReason accessed on Success"` instead of `"Unknown reason"` since there is no reason for error.

I double checked that everything from develop is there, I will be more careful next time!

angel-ayala (dismissed):
Yep, sounds good to me! Nice done!

TimonSchreiber:
Sorry @angel-ayala but I did not dismiss your review! Not sure what happended here...

angel-ayala (dismissed):
Not sure either

angel-ayala (approved):

angel-ayala (dismissed):
approved

angel-ayala:
something happened, now even approved is blocked

angel-ayala (approved):
approve

angel-ayala:
something odd happened, [workaround comment](https://github.com/orgs/community/discussions/58535#discussioncomment-6233887)

PR #49, opened 2026-04-12
refactor: method return types and number separators:

TimonSchreiber:
Minor style changes, no functional changes.

**Standardize use of method return types aligning with android style guide:**
Always define the return type unless it is `Unit` or very basic Boolean expression. If the method is a delegating call, always write the return type.

**Consisten use of numeric literal separators for numbers with 5 or more digits**


angel-ayala (approved):
Thanks @TimonSchreiber , this definitively enhances the code's quality and moves toward a standardized notation, that I must be sincere, I didn't know it. Good work.

PR #50, opened 2026-04-13
Fix init/shutdown sequence with split concerns

### LINKED ISSUE #46: Problem with boot/shutdown sequence
Something happened from last changes and I encountered problems in the boot and shutdown sequence. Probably something was broken with the new command pattern and the `monitorJob` to sync current state.
I will look up to fix this in the next days.

### LINKED ISSUE #47: Problems with WebRTC
WebRTC component is not working anymore, I don't have the exact log right now but is something related to the `mediaOptions` attribute that is being used before initialization.

angel-ayala:
Closes #46

The following changes move the `AircraftHandler` init/shutdown (I/S) sequence from `WenuLinkService` to `WenuLinkApp`, making now the application layer responsible of the start and finish contract with the SDK. In such way, a proper initialization is first made before consume the API features for different components, for example #47 appears to also be solved here. Nevertheless, I did not ran such test. This PR was only focused to the `AircraftHandler` I/S sequence. Another PR will be opened for other issues.

Addressed changes were mostly required to fit the I/S of the command pattern into a suitable way. Additionally, the `monitorJob` was also updated to ensure loops of desired `ms` and to perform non-suspendable sensor checks seems as an async method, is not required anymore.

One change that can be addressed in a near future is the moment where enforce or continuously ask for home position, maybe on the `TakeOffCommand` is a good place instead of iteratively check if was set.

### DIFF
--- a/app/src/main/java/org/WenuLink/WenuLinkApp.kt
+++ b/app/src/main/java/org/WenuLink/WenuLinkApp.kt
@@ -242,7 +242,10 @@ class WenuLinkApp : Application() {
     }
 
     fun disconnectAircraft() {
+        if (connectJob?.isActive == true) return
+
         if (!::wenuLinkHandler.isInitialized) return
+
         appScope.launch {
             val shutdownResult = wenuLinkHandler.unloadComponents(false)
             if (shutdownResult.hasError) {
--- a/app/src/main/java/org/WenuLink/WenuLinkApp.kt
+++ b/app/src/main/java/org/WenuLink/WenuLinkApp.kt
@@ -13,8 +13,6 @@ import androidx.core.content.ContextCompat
 import androidx.multidex.MultiDex
 import com.cySdkyc.clx.Helper
 import io.getstream.log.AndroidStreamLogger
-import io.getstream.log.StreamLog
-import io.getstream.log.TaggedLogger
 import io.getstream.log.taggedLogger
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -39,10 +37,12 @@ class WenuLinkApp : Application() {
     private var isContextAttached = false
     private var isConnectingAircraft = false
     var wenuLinkService: WenuLinkService? = null
+
     // Basic status reporting
     val isPermissionsGranted = MutableStateFlow(false)
     val workflowStatus = MutableStateFlow("Idle")
     val sdkStatus = MutableStateFlow("Idle")
+
     // SDK state handling
     val isRegistered = MutableStateFlow(false)
     val bindingString = MutableStateFlow("Idle")
@@ -207,7 +207,7 @@ class WenuLinkApp : Application() {
     private suspend fun waitComponentsAndStartHandler() {
         updateWorkflow("Waiting for components")
         val allComponentsPresent =
-            AsyncUtils.waitTimeout(100L, 30_000, APIManager::areComponentsPresent)
+            AsyncUtils.waitTimeout(100L, 30_000L, APIManager::areComponentsPresent)
 
         if (!allComponentsPresent) {
             updateWorkflow("No components found, try again...")
@@ -235,7 +235,9 @@ class WenuLinkApp : Application() {
             if (bootError == null) {
                 isAircraftBoot.value = true
                 updateWorkflow("Connected ready for service")
-            } else updateWorkflow("Boot error: $bootError")
+            } else {
+                updateWorkflow("Boot error: $bootError")
+            }
             isConnectingAircraft = false
         }
     }
@@ -245,8 +247,11 @@ class WenuLinkApp : Application() {
         appScope.launch {
             if (!::wenuLinkHandler.isInitialized) return@launch
             val shutdownError = wenuLinkHandler.unloadComponents(false)
-            if (shutdownError == null) isAircraftBoot.value = false
-            else logger.i { "Shutdown error: $shutdownError" }
+            if (shutdownError == null) {
+                isAircraftBoot.value = false
+            } else {
+                logger.i { "Shutdown error: $shutdownError" }
+            }
             wenuLinkHandler.enableSimulation(false)
         }
     }
@@ -255,9 +260,10 @@ class WenuLinkApp : Application() {
         logger.d { "onProductConnected $connected" }
         isAircraftPresent.value = connected
         if (connected) {
+            updateStatus("Aircraft present")
             initHandler()
         } else {
-            updateStatus("Aircraft disconnected")
+            updateStatus("No aircraft present")
             isSimulationReady.value = false
         }
     }
--- a/app/src/main/java/org/WenuLink/WenuLinkApp.kt
+++ b/app/src/main/java/org/WenuLink/WenuLinkApp.kt
@@ -34,10 +34,10 @@ class WenuLinkApp : Application() {
     }
     private val logger by taggedLogger(WenuLinkApp::class.java.simpleName)
     private val appScope = CoroutineScope(SupervisorJob() + Dispatchers.Main)
-    private var startingJob: Job? = null
+    private var initJob: Job? = null
+    private var connectJob: Job? = null
     private lateinit var wenuLinkHandler: WenuLinkHandler
     private var isContextAttached = false
-    private var isConnectingAircraft = false
     var wenuLinkService: WenuLinkService? = null
 
     // Basic status reporting
@@ -163,8 +163,6 @@ class WenuLinkApp : Application() {
             this::startService
         }
         startFunction(Intent(this, WenuLinkService::class.java))
-        wenuLinkService?.runServices()
-        isServiceUp.value = true
     }
 
     suspend fun stopWenulinkService() {
@@ -172,7 +170,6 @@ class WenuLinkApp : Application() {
 
         wenuLinkService?.terminate()
         stopService(Intent(this, WenuLinkService::class.java))
-        isServiceUp.value = false
     }
 
     private fun onRegistration(result: UnitResult) {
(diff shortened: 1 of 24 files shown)


TimonSchreiber (changes requested):
## Summary

The architectural direction here is sound. Moving the aircraft init/shutdown sequence from `WenuLinkService` into `WenuLinkApp` and establishing a clear three-stage boot (Aircraft/Components -> `WenuLinkService` -> MAVLink/WebRTC) resolves the previous tight coupling that caused issue #46. The `monitorJob` improvements, the synchronous sensor check refactor, the `suspendCancellableCoroutine` conversions in `SimManager`, the `FlightModeTransition` unification, and the metadata retrieval consolidation in `SDKUtils` are all solid work.

---

## Bugs

### 1. `launchWenulinkService()`: race condition and premature state flag

Location: `WenuLinkApp.kt`

```kotlin
fun launchWenulinkService() {
    // ...
    startFunction(Intent(this, WenuLinkService::class.java))
    wenuLinkService?.runServices()  // almost certainly null at this point
    isServiceUp.value = true        // set before the service is actually running
}
```

`startForegroundService` / `startService` are asynchronous, `wenuLinkService` is assigned via a later callback. At the point these two lines execute, the service has not started, `wenuLinkService?.runServices()` does nothing and `isServiceUp` is set to `true` prematurely.

The old code had at least a `waitReady` delay.

A better place for both is the service `WenuLinkService.onStartCommand()`:

```kotlin
override fun onStartCommand() {
    // ... existing setup
    startForegroundServiceWithNotification()
    runServices()
    thisApp.isServiceUp.value = true
    return START_STICKY
}
```

---

### 2. `connectAircraft()`: string template bug

Location: `WenuLinkApp.kt`

```kotlin
updateWorkflow("Boot error: $bootError.reason")
```

Should be

Fix:

```kotlin
updateWorkflow("Boot error: ${bootResult.reason}")
```

---

### 3. `monitorJob`: `CancellationException` is missing

Location: `WenuLinkHandler.kt`, `startMonitorJob()`

```kotlin
try {
    // ...
} catch (e: CancellationException) {
    throw e
} catch (e: Exception) {
    logger.e { "Monitor loop error: ${e.message}" }
}
```

---

### 4. `apiDestroy()` unguarded `lateinit` access

Location: `WenuLinkApp.kt`

```kotlin
fun apiDestroy() {
    if (::wenuLinkHandler.isInitialized) wenuLinkHandler.unload()
    appScope.cancel()
    APIManager.destroy()
}
```

---

### 5. `FlightModeTransition`: GUIDED restriction silently dropped

Location: `AircraftState.kt`

Old `isModeAllowed()`:

```kotlin
ArduCopterFlightMode.GUIDED -> state.isFlying() || state.isStandBy()
```

New `FlightModeTransition.canTransition()`:

```kotlin
ArduCopterFlightMode.LAND, ArduCopterFlightMode.RTL -> flyingMode(from)
else -> CommandResult.ok   // GUIDED falls here, always allowed from any state
```

GUIDED can now be requested from any aircraft state now.

```kotlin
ArduCopterFlightMode.GUIDED if (from.isFlying() || from.isStandBy()) -> CommandResult.ok
ArduCopterFlightMode.GUIDED ->
    CommandResult.error("GUIDED not allowed from ${from.flightMode}")
ArduCopterFlightMode.LAND, ArduCopterFlightMode.RTL -> flyingMode(from)
else -> CommandResult.ok
```

---

### 6. `MainActivity`: button label operator precedence

Location: `MainActivity.kt`

```kotlin
Text(if (!isAircraftUp) "Connect" else "Disconnect" + " aircraft")
```

Will result in either `Connect` or `Disconnect aircraft`.

Fix:

```kotlin
Text("${if (!isAircraftUp) "Connect" else "Disconnect"} aircraft")
```

---

## Design Observations

### `coroutineScope { }.join()` is redundant

Location: `WenuLinkHandler.kt`, `startMonitorJob()`

Calling `.join()` on an already-completed job is meaningless

---

### `isConnectingAircraft` guard is checked inside the coroutine, not before it

Location: `WenuLinkApp.kt`, `connectAircraft()` and `disconnectAircraft()`

```kotlin
fun connectAircraft(useSimulation: Boolean = false) {
    if (isConnectingAircraft) return  // checked on calling thread
    appScope.launch {
        if (!::wenuLinkHandler.isInitialized) return@launch
        isConnectingAircraft = true   // set after launch
        // ...
    }
}
```

The check and the assignment are not atomic. If `connectAircraft()` is called twice before the first coroutine's `isConnectingAircraft = true` line has executed, both calls will pass the guard and launch concurrent connect sequences.

Compare with `initHandler()`, which correctly checks `startingJob != null` on the calling thread before launching. The same pattern could be applied here, for example by using a `Job?` field as the guard rather than a plain `Boolean`.

```kotlin
private var connectJob: Job? = null

fun connectAircraft(useSimulation: Boolean = false) {
    if (connectJob?.isActive == true) return
    connectJob = appScope.launch {
        // ...
    }
}
```

---

### Simulation is no longer disabled on service stop

Location: `WenuLinkApp.kt`, `stopWenulinkService()`

The old implementation called `wenuLinkHandler?.enableSimulation(false)` on service stop. This is now gone. `disconnectAircraft()` does call `enableSimulation(false)`, but only after `unloadComponents()` completes. If someone calls `stopWenulinkService()` without also calling `disconnectAircraft()`, simulation remains active. Is this intentional?

---

### `unloadComponents()` failure is not surfaced in the UI

Location: `WenuLinkApp.kt`, `disconnectAircraft()`

```kotlin
val shutdownError = wenuLinkHandler.unloadComponents(false)
if (shutdownError is CommandResult.Failure) {
    logger.i { "Shutdown error: ${shutdownError.reason}" }
} else {
    isAircraftBoot.value = false
}
```

A failed shutdown leaves the aircraft in an indeterminate state. The user only sees the aircraft remain listed as connected, with no explanation. At a minimum, a `updateWorkflow("Disconnect error: ${shutdownError.reason}")` call would help diagnose the problem during testing. The log level also seems off, this is an error, not an info-level event.

Generally I think the correct suffix for variables of type `CommandResult` should be `Result`, so `shutdownResult` would be better.

angel-ayala:
Great! I just updated the code and addressed almost all the changes with the following exceptions.
Bug .3 I don't see any specific need to have it since as it is, already handle the wide spectrum of Exceptions. Regards to Bug .5 I that was intentional, and right now avoids to call `enforceModeConsistency` since I don't see an specific rule set to check custom flight mode transition. For example, in the Ardupilot firmware, you must be in `GUIDED` mode to do a takeoff, with som other specific constraint when must be capable to arm.

Regards the simulation, is an especial case where if the user choose to enable the simulation feature, this operation is not being fully reverted, requiring to reboot the Aircraft to switch back and connect to real hardware. No other test has been done, possibly an application restart also solves it. Nevertheless, right now does not seems to be a bug, is more related to the initialization process and the positioning system internals. At least current test with the RC and Aircraft already connected have successful results in both modes for init/shutdown sequeces.

Any other related bug or issue I think can be address in a different PR.

app/src/main/java/org/WenuLink/WenuLinkApp.kt
```diff
@@ -129,55 +147,122 @@ class WenuLinkApp : Application() {
...
+                updateWorkflow("Boot error: ${bootResult.errorReason}")
+            } else {
+                isAircraftBoot.value = true
+                updateWorkflow("Connected and ready for services")
+            }
+        }
+    }
+
+    fun disconnectAircraft() {
+        if (!::wenuLinkHandler.isInitialized) return
```
TimonSchreiber, line 247:
```suggestion
        if (connectJob?.isActive == true) return

        if (!::wenuLinkHandler.isInitialized) return

```

app/src/main/java/org/WenuLink/adapters/WenuLinkHandler.kt
```diff
@@ -106,38 +98,56 @@ class WenuLinkHandler : CommandHandler<WenuLinkHandler>() {
...
+                            cameraHooks()
+                        }
 
-                    launch {
-                        safetyChecks()
+                        launch {
+                            safetyChecks()
+                        }
                     }
                 } catch (e: Exception) {
```
TimonSchreiber, line 134:
```suggestion
                } catch (e: CancellationException) {
                    throw e
                } catch (e: Exception) {
```

TimonSchreiber (changes requested):
You are correct on both points. Bug 5 was only mentioned because it was not obvious to me that the omission of `GUIDED` was intentional.

While I see that adding `CancellationException` to the `monitorJob` is not strictly necessary, it would prevent us from logging an error: "Monitor loop error: ...", just because the monitorJob was cancelled, which is an expected behaviour. So while it works this way, it also pollutes the log output with an invalid error message. The correct way  is to catch the `CancellationException` explicitly instead of handling it the same as every other exception.

There is only one minor thing that should change before merging: the `disconnectAircraft` needs to have the same guards as `connectAircraft`, see inline suggestion.

Everything else looks good!

angel-ayala:
Oh I see, makes a lot of sense the `CancellationException` catch now, but instead of throwing the exception I just print a warning log with such situation, seems to be inappropriate to launch an exception on a try-catch block.

Now everything should be fine.

TimonSchreiber (dismissed):
It appears that throwing the `CancellationExpection` in some form or another is important, but it is not necessary to use `throw e`:
https://stackoverflow.com/questions/76259793/is-it-necceary-to-rethrow-the-cancellationexception-in-kotlin

Would you consider using `ensureActive`? I think that can even be used in the `Exception`-catch before the logger and then the `CancellationException`-catch might not be necessary after all. Sorry for the confusion.

Either way I approve this pr, good job!

angel-ayala:
I just read the link and now I see your point, definitively I will update it. I just checked and apparently I cannot use `if (e is CancellationException)`, Android Studio is saying that such condition is always `false`. So I just put the `currentCoroutineContext().ensureActive()` on `catch (e: CancellationException)`

TimonSchreiber (approved):
Lgtm

PR #54, opened 2026-04-13
Fix decoding of param4 (absolute/relative)

TimonSchreiber:
The most relevant change from pr #43 , which solves a bug with relative/absolute offset.


angel-ayala (approved):
good catch @TimonSchreiber !

PR #58, opened 2026-04-14
Fix: handle NaN values in mapAction to prevent crash (Issue #45)

EIHEI2:
## Summary
This PR fixes the 'Cannot round NaN value' crash that occurs when uploading missions with simulation feature.

## Root Cause
The mapAction function in MissionManager.kt was calling roundToInt() on angle values without checking for NaN first. When angle values are NaN, Kotlin's roundToInt() throws IllegalArgumentException.

## Fix
Added NaN checks before calling roundToInt() for both RotateAction and GimbalPitchAction:
- RotateAction: if (action.angle.isNaN()) 0 else action.angle.roundToInt()
- GimbalPitchAction: if (action.angle.isNaN()) 0 else action.angle.roundToInt()

## Testing
The fix prevents the crash by defaulting to 0 when angle is NaN.

Closes #45


angel-ayala:
Hi @EIHEI2 thanks for your contribution, please use `develop` as target branch instead `main`. Can please also adds which commit did you used as base for such changes. Thanks.

angel-ayala (changes requested):
update branch target to `develop`

angel-ayala:
@EIHEI2 seems to be an agent bot, if there is no updates in the following days, I will close this PR due to inactivity.

TimonSchreiber:
@angel-ayala I thinks it's time to remove this pr

angel-ayala:
Closing due to inactivity

PR #59, opened 2026-04-14
GPS fix type mapping and Message Construction

TimonSchreiber:
## Refactor message construction and fix GPS fix type mapping

The `TelemetryData` class held a `gpsLevel: List<Boolean>` field, a one-hot encoding of DJI's `GPSSignalLevel` enum, alongside a `gpsFixType: Int` field that was always defaulted to `0` and never populated. `msgRawGPSInt` then had to extract that value to recover the original level, and its mapping was incomplete: `DJI's GPSSignalLevel` has 12 values (LEVEL_0–LEVEL_10 + NONE) but only levels 0–5 were handled, leaving the rest incorrectly falling through to `GPS_FIX_TYPE_NO_FIX`.

This PR fixes the root cause by doing the DJI -> MAVLink mapping at the SDK boundary, where the DJI type is already available, and storing only the resolved MAVLink fix type in `TelemetryData`:

- Add `GPSMapper.toMavlinkFixType(GPSSignalLevel)` in the adapter layer (alongside `BatteryMapper`), with a complete mapping covering all 12 DJI levels.
- Replace `gpsLevel: List<Boolean>` with `gpsFixType: Int` (non-defaulting) in `TelemetryData`, removing the dead default-zero field.
- Update `FCManager` and `SimManager` to call `GPSMapper` at construction time; delete `SDKUtils.gpsSignalLevelFlags`.
- `msgRawGPSInt` now simply reads `telemetryData.gpsFixType.toShort()`.

The remaining changes apply the same `.apply {}`/expression-body pattern to all other message-construction functions across `ConnectionController`, `NavigationController`, `CommandController`, `CameraController`, and `MessageUtils`, eliminating the intermediate` val msg =` /` return msg` boilerplate throughout.

### DIFF
--- a/app/src/main/java/org/WenuLink/mavlink/controllers/NavigationController.kt
+++ b/app/src/main/java/org/WenuLink/mavlink/controllers/NavigationController.kt
@@ -357,20 +357,11 @@ class NavigationController(
             return@apply
         }
 
-        val gpsLevel = telemetryData.gpsLevel.indexOfFirst { it }
-
         time_usec = MessageUtils.getMicroTime()
         lat = MessageUtils.coordinateDJI2MAVLink(telemetryData.latitude)
         lon = MessageUtils.coordinateDJI2MAVLink(telemetryData.longitude)
         satellites_visible = telemetryData.satelliteCount.toShort()
-
-        // DJI reports signal quality as a level 0-11; mapped to Mavlink  fix types below.
-        fix_type = when (gpsLevel) {
-            0, 1 -> GPS_FIX_TYPE.GPS_FIX_TYPE_NO_FIX.toShort()
-            2 -> GPS_FIX_TYPE.GPS_FIX_TYPE_2D_FIX.toShort()
-            3, 4, 5 -> GPS_FIX_TYPE.GPS_FIX_TYPE_3D_FIX.toShort()
-            else -> GPS_FIX_TYPE.GPS_FIX_TYPE_NO_FIX.toShort()
-        }
+        fix_type = telemetryData.gpsFixType.toShort()
     }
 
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/WenuLink/mavlink/controllers/NavigationController.kt
```diff
@@ -311,96 +312,76 @@ class NavigationController(
...
-            msg.fix_type = GPS_FIX_TYPE.GPS_FIX_TYPE_2D_FIX.toShort()
-        } else if (
-            telemetryData.gpsLevel[3] ||
-            telemetryData.gpsLevel[4] ||
-            telemetryData.gpsLevel[5]
-        ) {
-            msg.fix_type = GPS_FIX_TYPE.GPS_FIX_TYPE_3D_FIX.toShort()
-        } else {
-            msg.fix_type = GPS_FIX_TYPE.GPS_FIX_TYPE_NO_FIX.toShort()
+            fix_type = GPS_FIX_TYPE.GPS_FIX_TYPE_NO_GPS.toShort()
```
angel-ayala:
```suggestion
            fix_type = GPS_FIX_TYPE.GPS_FIX_TYPE_2D_FIX.toShort()
```

I was wondering why I defined simulation with this line, but I think is better to update this since the simulation feature is always delivering position information.

angel-ayala (dismissed):
Great work @TimonSchreiber as usual, presented changes are a very good improvement. I just would like to ask to update the GPS fix type for the Simulation feature case, I'm not sure why I defined as it is, but is better to update it.
I will approve this PR because that can also be updated later.

TimonSchreiber:
Good call on that one @angel-ayala I had missed adjusting the fix type for simulation.

angel-ayala (approved):
Cool! It is even better!

PR #64, opened 2026-04-16
Fix missing CommandResult checks

### LINKED ISSUE #57: Update missing return of CommandResult
I was looking into the code and there are still some missing updates to fully implement the CommanResult return type, such as the case of:

https://github.com/WenuLink/wenu-link-android/blob/0f3e2110ce870e383a9eddb958065debaf32000f/app/src/main/java/org/WenuLink/mavlink/controllers/CameraController.kt#L302-L308

A more exhaustive search must be done to update all ocurrences. Maybe by searching `error != null` or `error == null` could help find most of them.

TimonSchreiber:
Found three cases where the lambda didn't match the new `CommandResult` type. All other cases of `error == null` or `error != null` are on the manager classes or DJI classes and reflect the `String?` type the DJI SDK uses for errors.

If we want to define a result type for the manager classes it should not also be `CommandResult` since the name would be inappropriate.

Let me know if I missed anything.

Closes #57

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
+++ b/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
@@ -21,8 +21,6 @@ data class BootCommand(val timeout: Long = 5000L) : AircraftCommand {
         // TODO: check if compatible with CancellableCoroutine
         ctx.dispatchTransition(BootTransition)
         return ctx.boot(timeout)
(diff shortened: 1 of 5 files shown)


angel-ayala (approved):
yeah makes sense to me in using another Result type for the SDK manager classes, I will update the issue related to IManager improvement, looking to discuss all related changes to the SDK interfacing classes in the same place.

LGTM

PR #73, opened 2026-04-20
Fix new line and remove reassignment

TimonSchreiber:
Minor fix on the notification content text


angel-ayala (approved):
LGTM, I'm not sure if it is the adequate fix since still using `var` and the notification bar appears to not be updated when a service is initialized or terminated, but probably we can discuss that looking over the entire UI/UX topic.

PR #62, opened 2026-04-15
Fix RequestTakeoff command

angel-ayala:
These changes updated the aircraft state transition to correctly advance from boot to isFlying, some validations and restriction were intentionally leave out, since there were not enough reasons to have it considering special flightmode cases

Fixes #56 and `SetHomePositionCommand` with a more adequate logic

PR #74, opened 2026-04-20
Replace null checks

TimonSchreiber:
Replaces repeated null checks with a more kotlin-esque `?.let { } ?:` expression similar to changes in #48 and #62

The base for this refactor was the `TelemetryHandler.listenSimulation` method which I considered in need of a rewrite to make the intent clearer


app/src/main/java/org/WenuLink/mavlink/controllers/CameraController.kt
```diff
@@ -161,16 +161,13 @@ class CameraController(
         }
     )
 
-    private fun getCamera(targetCamera: Int): CameraMetadata? {
-        val cameraInfo = handler.camera.getCamera(targetCamera)
-        if (cameraInfo == null) {
-            logger.d { "Camera index $targetCamera not found" }
+    private fun getCamera(targetCamera: Int): CameraMetadata? =
+        handler.camera.getCamera(targetCamera).also {
+            if (it == null) logger.d { "Camera index $targetCamera not found" }
```
angel-ayala: I think there should be another way to avoid this null check

app/src/main/java/org/WenuLink/mavlink/controllers/CommandController.kt
```diff
@@ -113,18 +113,15 @@ class CommandController(override var client: MAVLinkClient, override val handler
...
-        }
+        ArduCopterFlightMode.from(params.customMode)
+            ?.let {
+                if (handler.aircraft.requestMode(it).isOk) {
+                    sendCommandAck(commandMsg.command, MAV_RESULT.MAV_RESULT_ACCEPTED)
+                } else {
+                    sendCommandAck(commandMsg.command, MAV_RESULT.MAV_RESULT_DENIED)
+                }
+            }
+            ?: sendCommandAck(commandMsg.command, MAV_RESULT.MAV_RESULT_DENIED)
```
angel-ayala: The log line should be split. One inside the `?.let {}` block to print the log with the proper name `logger.d { "FlightMode requested: ${it}" }`. And the other after the elvis operator `?:` indicating `logger.d { "FlightMode ${params.customMode} not found" }`

angel-ayala (changes requested):
Nice done @TimonSchreiber ! Only a few minors were you still comparing `null` and the flight mode request log to have and easy-to-read log.

TimonSchreiber:
Good catch @angel-ayala , let me know if you find any more null checks that should be replaced (mostly after method returns)

angel-ayala (dismissed):
from the proposed changes, those were all! If something else appears during my changes, I will try to apply too.
Buen trabajo!

angel-ayala:
there is a diff file in the commit

TimonSchreiber:
Sorry about that @angel-ayala , that was part of my ui work and it didn't get stashed when switching to this feature branch

angel-ayala (approved):
no worries

PR #76, opened 2026-04-20
refactor: return UnitResult from dispatchControlAuthority

TimonSchreiber:
Align `dispatchControlAuthority` with the rest of the handler layer by returning `UnitResult` instead of `Unit`, and make the `RequestCommand` call sites observe the outcome.

This is based on the changes from #62 and should be merged after it

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/WenuLinkHandler.kt
+++ b/app/src/main/java/org/WenuLink/adapters/WenuLinkHandler.kt
@@ -28,6 +28,7 @@ import org.WenuLink.adapters.mission.ResumeActionCommand
 import org.WenuLink.adapters.mission.ResumeWaypointMission
 import org.WenuLink.adapters.mission.StopWaypointMission
 import org.WenuLink.commands.CommandHandler
+import org.WenuLink.commands.CommandResult
 import org.WenuLink.commands.UnitResult
 
 class WenuLinkHandler : CommandHandler<WenuLinkHandler>() {
@@ -174,6 +175,7 @@ class WenuLinkHandler : CommandHandler<WenuLinkHandler>() {
     suspend fun loadComponents(scope: CoroutineScope): UnitResult {
         // prevent cancel command when manualControl() after boot
         dispatchControlAuthority(ControlAuthorityType.REMOTE_CONTROLLER)
+
         val bootResult = dispatchAndAwait(WenuLinkCommand.Aircraft(BootCommand(30_000L)))
         if (bootResult.isOk) {
             manualControl()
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/WenuLink/adapters/WenuLinkHandler.kt
```diff
@@ -194,23 +197,28 @@ class WenuLinkHandler : CommandHandler<WenuLinkHandler>() {
...
-    fun dispatchControlAuthority(authority: ControlAuthorityType) {
+    fun dispatchControlAuthority(authority: ControlAuthorityType): UnitResult {
         // Decide policy: reject or stop mission
-        if (!controlAuthority.isNewAuthority(authority)) return
+        if (!controlAuthority.isNewAuthority(authority)) return CommandResult.ok
+
+        if (controlAuthority.isRemote()) {
+            logger.d { "Skip authority switch: remote active" }
+            return CommandResult.error("Remote pilot has authority")
+        }
```
angel-ayala, line 215: I understand that this is an extra-safety measure, however its implementation will break up all processes, since after boot it already calls `WenuLinkHandler.manualControl()` and will always keep it returning error.

angel-ayala (changes requested):
I agree with the UnitResult return, however, adding the extra-safety measure to keep the `controlAuthority` always `isRemote()` is dangerous since there is no other way to disable it. I suggest to open a Discussion page or issue to discuss a more proper logic if you see it necessary. For now I will leave it as it is.

A possible upgrade could be to listen for a certain RC button to lock/unlock transitions from `REMOTE_CONTROLLER` authority.

app/src/main/java/org/WenuLink/adapters/WenuLinkHandler.kt
angel-ayala: I just noticed that I added this bug earlier, aiming to fix it in #80

TimonSchreiber:
I saw the `isRemote` guard in your pr #79 @angel-ayala and added it as well, splitting the predicate to log the remote case. I probably should have based this branch on `fix/do-reposition`, but wasn't sure what would have happened if this pr merges before #79 . Do you want to go forward with the return type change and use the fix from #80 or leave this pr out for now and come back after #80 ?

angel-ayala:
since #79 was already merged and #80 still need to rebase develop, please fix the conflicts to merge this first and then I update #80 with this PR changes.

TimonSchreiber:
Please check the `RequestMissionAction` again to make sure that `ControlAuthorityType.TIMELINE_COMMAND` is the correct type, the merge conflict was a bit confusing

angel-ayala (approved):
yes, that's correct `MissionActionCommand` are interfacing the SDK's `MissionControl` with `TimelineElement`s thats why of being a `TIMELINE_COMMAND`.

Be careful with only using rebase, there has been situations , for example now when I rebase `fix/do-reposition` from `fix/upload-mission` branch I lost certain changes already made in `fix/upload-mission` maybe is something related to how old it is, not sure, that's why I will do merge instead.

PR #77, opened 2026-04-20
Use ImageStartCaptureMessage in requestStartCapture

TimonSchreiber:


angel-ayala (approved):
Good catch! I only wonder if `intervalSec` is always used as `Int`, is not better to declare it as `Int` instead `Float`?

TimonSchreiber:
Only the dji sdk expects the interval seconds to be integers, mavlink treats them as floats. I don't think it is a good idea to move that conversion to the message layer, even if the only use of that value is as an integer

PR #79, opened 2026-04-20
Fix RequestReposition

### LINKED ISSUE #72: Problems with do reposition
`MAV_CMD_DO_REPOSITION` command was not working and errors silently omitted. Main issue is related to the default flight speed defined on the action side requiring to be in [2, 15]. Additionally, current implementation miss the use of parameters translation and handling, another reason for taking care of #71 .

I already addressed this, waiting #62 to be merged to avoid inumerous changes since it was done on top of it.

angel-ayala:
Fixes #72 by correcting default flight speed and other mission related reorganization.

The `RepositionAction` was only working for takeoff but no on request of DO_REPOSITION, with a main bug related to the default flight speed = 1, below the minimum required by the SDK (2m/s). Since the logger function is not reachable by any command, auxiliary methods were created on MissionHandler making it easier to follow.

Fixes MAV_CMD_DO_REPOSITION parameters processing as required by SDK.

In this changes I aimed to keep all manager classes accessible by handlers but no by commands, but then I saw that this is not consistent, in certain cases, commands are still calling SDK manager related methods. Probably we should also define a way an keep it alongside the development.
Specifically here, `MissionHandler` now import the SDK's `MissionAction` class violating the restriction of calling SDK dependencies only inside Manager classes, requiring a major refactor to avoid such situation.

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/RequestCommand.kt
+++ b/app/src/main/java/org/WenuLink/adapters/RequestCommand.kt
@@ -125,7 +125,7 @@ open class RequestMissionAction(private val action: MissionActionCommand) :
     override suspend fun execute(ctx: WenuLinkHandler): UnitResult {
         super.execute(ctx)
 
-        ctx.dispatchControlAuthority(ControlAuthorityType.WAYPOINT_MISSION)
+        ctx.dispatchControlAuthority(ControlAuthorityType.TIMELINE_COMMAND)
 
         return ctx.dispatchAndAwait(WenuLinkCommand.Mission(action))
     }
--- a/app/src/main/java/org/WenuLink/adapters/mission/MissionCommands.kt
+++ b/app/src/main/java/org/WenuLink/adapters/mission/MissionCommands.kt
@@ -169,7 +169,7 @@ interface MissionActionCommand : MissionCommand {
         else -> CommandResult.ok
     }
 
-    override suspend fun onStop(ctx: MissionHandler) = ctx.onActionError("Stop invoked")
+    override suspend fun onStop(ctx: MissionHandler) = ctx.stopAction("Stop invoked")
 }
 
 data class DelayAction(val timeMillis: Long) : MissionActionCommand {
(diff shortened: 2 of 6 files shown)


TimonSchreiber (changes requested):
Great fix @angel-ayala , I just have some questions and comments.

### Major

#### 1. Mission `cont.isActive` guards
The old `ActionCommand.execute` wrapped both resume sites in a guard, this could possibly result in an `IllegalStateException`.

### 2. `lastActionKey` as a shared mutable state of the singleton `MissionHandler`
If a second action is scheduled before the first one's callback fires, `lastActionKey` gets overwritten, and the first action's cleanup path will `removeCallback` on the wrong key. Maybe each scheduled action could have their own handle, hand that back to the caller and have teardown take it as a parameter: `ctx.teardownAction(handle)`.

---

### Minor

#### 3. `onActionError` is misnamed
It runs on stop and on cancellation, and then logs at ERROR-level, meaning every normal stop/cacel gets treates like an error. Maybe rename to `teardownAction` and change the log level to information.

#### 4. Clamping speed in `RepositionAction`
The clamping is a good idea, but could benefit from a log statement similar to `MissionHandler.setSpeed`, so every out-of-range speed gets tracked.
Clamping could also happen in `DoRepositionCommandInt` so that `RepositionAction` already gets a valid speed at construction.

#### 5. `onActionFinish` callback signature
It is defined as `onFinish: (UnitResult) -> Unit)`, but the only call hardcodes `CommandResult.ok`. The signature could be simplified to `( ) -> Unit`.

#### 6. Naming of `scheduleUrgentAction`
"Urgent" implies priority/preemption that isn't really there, the intention is just "clear the timeline then schedule this". `replaceSchedule(action)` or `scheduleImmediateAction` would read more accurately

angel-ayala:
Nice review! I did not apply too much changes regards the major, specifically for 1. I think that we should review the entire project to decide this of possible during tests since there are a lot of times that such guard is not present, and so far everything is working. That kind of guard was intentionally put in the `WenuLinkHandler.dispatchAndAwait()`, i have added before in commands, but after removal nothing remarkable happens so I keep it simple.


Regards the minors, all of them were addressed now improving readability.

TimonSchreiber (approved):
If leaving out the `isActive` checks causes no problem that is fine with me, I was just trying to be cautious.

This looks good now

PR #82, opened 2026-04-21
Rework UI architecture and introduce Jetpack Compose navigation

### LINKED ISSUE #4: Improve GUI

TimonSchreiber:
Picks up from #29 with permission from @ElitonSena and @angel-ayala . All credit for the original implementation goes to @ElitonSena .

This PR applies the changes requested in the review of #29 and rebases the branch onto the current state of `develop`.

#### Changes compared to #29 :
- Rebase onto `develop`; resolve merge conflicts in `MainActivity.kt`
- Updated `MainActivity` and `AppNavigation` to remove the now deleted `HomeViewModel` dependency and observe state directly from `WenuLinkApp`
- Replace `sdkStatus.contains("Connected")` with `app.isAircraftBoot` for the `canRunService` flag
- Moved `navigation-compose` and `material-icons-extended` dependencies into `libs.versions.toml`
- Update contributor description in `AboutScreen` and `README`
- Applied ktlint formatting and optimized wildcard imports

#### Still to do before merge:
- [x] Remove or rework `KeyScreen`
- [x] Change IP fields to store full addresses including port (e.g. `192.168.1.220:14550`)
- [x] Split IP settings into separate MAVLink and WebRTC screens
- [x] Rename `PrefsManager` to `WenuLinkPreferences` and move to `org.WenuLink`

#### Notes:
- `isDataFlowing` and telemetry summary fields were removed since they are no longer exposed via `TelemetryHandler`

---

Closes #29
Relates to #4


angel-ayala:
Great!

Two main suggestions to add in the roadmap:
- A widget (I don't know a technical name for it) or some UI element as a checkbox to enable/disable simulation if availabe.
- Present a log space with lines called from `logger.i` maybe or some other rule set, or even a new function to show important lines to the user.

Possible the last point is better to address it in other PR, but things can be planned looking to implement everything, but only addressed after that. Let me know.

TimonSchreiber:
I agree with your suggestion, but would like to keep the ui changes for this pr close to the version @ElitonSena made and only work on the changes requested in the reviews.
All other UI requests and ideas should be discussed and implemented in separate prs

angel-ayala:
Changes seems fine, however is unfeasible to do any test with current stage, since there are no buttons triggering the initialization sequence which make impossible to check if settings changes are being reflected in the connection logic.
I think that such case, adding buttons to the UI still fits in fixing the UI rework. That was something that none of us mention it earlier, probably by lack of test intentions. If you check the old `MainActivity` you will see the buttons and conditions for each of them to appear in the UI.

Everything else sounds good.

angel-ayala (changes requested):
Please add the buttons to properly test settings update.

angel-ayala (changes requested):
The address in the UI part when define a new value setting is persisted when the app is closed and open again, I just changed one bit group. However, such value was not set when the service is loaded. Both MAVLink and WebRTC services keep trying connection to the hardcoded value.

```plain
(main:2) Creating MAVLinkClient for udp://192.168.1.220:14550.
(main:2) Connecting WebRTC client to ws://192.168.1.220:8090
```

Possibly the setting value must be read and set through the update method on the service class.

TimonSchreiber:
I completely missed updating the addresses inside `MAVLinkService` and `WebRTCService`, good catch @angel-ayala

angel-ayala (approved):
Great! I just test it and is working fine. I will accept this PR but I keep thinking that `ServiceAddress` update must happen on `startWebRTC` and `startMAVLinkService` before the client start, since the UI allows to stop and start those services independently, allowing to set the new address after stopping the service. If for some reason to achieve this must do more than changing the position of two lines, is preferably to open an issue describing this.

PR #86, opened 2026-04-23
Fix YUV color-format routing and MediaFormat lookups

TimonSchreiber:
Semi-planar formats were incorrectly routed through the I420 path.
These formats use interleaved chroma (typically NV12/NV21 layouts), while I420 expects separate U and V planes. As a result, the U plane was read correctly but the V plane was corrupted.

Route `COLOR_FormatYUV420SemiPlanar` and `COLOR_FormatYUV420PackedSemiPlanar` to the NV12 path, and keep planar formats on the I420 path.

Additional improvements:
- Use safe fallbacks for `KEY_STRIDE` and `KEY_SLICE_HEIGHT` (defaulting to width/height), since `MediaFormat.getInteger()` throws if keys are absent and they are not guaranteed to be present.
- Suppress deprecation warnings at function scope

### DIFF
--- a/app/src/main/java/org/WenuLink/webrtc/utils/VideoCapturerUtils.kt
+++ b/app/src/main/java/org/WenuLink/webrtc/utils/VideoCapturerUtils.kt
@@ -9,32 +9,33 @@ import org.webrtc.JavaI420Buffer
 import org.webrtc.NV12Buffer
 import org.webrtc.VideoFrame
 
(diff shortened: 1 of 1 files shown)


angel-ayala (approved):
Interesting, when I developed this part I just leave it as it works with the P4P, now seems to be more robust. I just tested it and its ok. I recommend to merge this before #82

Issue #83, opened 2026-04-21
[Bug]: Missing action commands during mission upload


angel-ayala:
### WenuLink App Version

ae1d862

### Drone Model

Phantom 4 Pro

### Test Environment

DJI Simulator

### Android Device (running WenuLink)

Xiaomi Redmi Note 10s

### Android Version

Android 13

### Remote Controller (if separate)

GL300F

### GCS Software & Version

QGroundControl v5.0.8 64 bit

### Related MAVLink Microservice

- [ ] Heartbeat/Connection
- [x] Mission
- [ ] Parameter
- [x] Command
- [ ] Camera
- [ ] Other (describe below)

### Steps to Reproduce

1. Launch app
2. Connect the drone using the simulation feature
3. Start services
4. Load a mission with 5 waypoints using delay or yaw actions in some of them

### Reproducibility

Always

### Expected Behavior

Upload the mission and start for a full execution

### Actual Behavior

Error during mission upload due to waypoints with actions

### Logs

[full_flight_plan.log](https://github.com/user-attachments/files/26945855/full_flight_plan.log)

### Additional Notes

I run a simple mission with no action in any waypoint and was ok, mission was uploaded and executed with no problems, so definitively is something related to processing actions. Possibly, actions are being considered as extra waypoint when must not give the reiterative requesting of apparently missing mission items on the logs

angel-ayala:
With changes from #90 I was able to identify the problem, and in my flight plan there I though that I was changing the orientation, but actually I tried an unsupported command that moves the gimbal, the logs I found `Unsupported mission command: 205` where 205 is the ID for `DO_MOUNT_CONTROL`, which is not currently implemented. So basically the bug here is not the process itself but the lack of required commands processing during mission upload.

angel-ayala:
I'm closing it, since we don't any plan yet to add other commands.

PR #90, opened 2026-04-23
Fix mission upload ACK protocol for non-waypoint items

TimonSchreiber:
- adds a separate sequence counter to `NavigationController`
- adds Message ACKs to all failure branches in `processMissionItem`
- adds check if `addWaypointNode` was successfull
- adds `MAV_CMD_CONDITION_DELAY`


angel-ayala (approved):
Changes are sound, possibly related with #83 , after merging I will re do the test

PR #91, opened 2026-04-27
Concurrency bug in `ParameterRegistry.filterAvailable`

TimonSchreiber:
`ParameterRegistry.load()` intermittently crashes with an NPE during `BootCommand` execution. The failure is timing-dependent and not always reproducible on the same hardware.

#### Stack trace

```
java.lang.NullPointerException: Attempt to invoke virtual method
  'org.WenuLink.parameters.ParameterSpec org.WenuLink.parameters.WenuLinkParameter.getSpec()'
  on a null object reference
    at org.WenuLink.parameters.ParameterRegistry.load(ParameterBase.kt:111)
    at org.WenuLink.parameters.ParameterRegistry$load$1.invokeSuspend(Unknown Source:14)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(...)
    at kotlinx.coroutines.android.HandlerContext$scheduleResumeAfterDelay$$inlined$Runnable$1.run(...)
```

Logcat context just before the crash:

```
AircraftHandler  Executing: BootCommand
AircraftHandler  Aircraft booting...
AircraftHandler  Loading parameters
AircraftHandler  Command failed: BootCommand
WenuLinkApp      Workflow: Boot error: ...getSpec() on a null object reference
AndroidRuntime   FATAL EXCEPTION: main
```

#### Root cause

`filterAvailable` mutates `availableParams`, `index`, and `counter` from DJI SDK callbacks that fire on background worker threads, with no synchronization:

```kotlin
parameters.forEach { spec ->
    spec.read { value ->        // DJI callback, background thread
        if (value != null) {
            spec.index = index
            index += 1
            availableParams += WenuLinkParameter(spec, value)
        }
        counter += 1
    }
}
```

Concurrent `+=` on a plain `ArrayList` can leave null elements in the backing array (one thread bumps `size`, another thread's element write hasn't been published yet). `load()` then iterates `params` via `associateBy { it.spec.index }` and dereferences a null slot, producing the NPE on `getSpec()`.

#### Solution

Replace the fire-and-poll design with structured concurrency: wrap each callback-based read in `suspendCoroutine`, run them under `coroutineScope { ... async { ... } }.awaitAll()`, and assign `spec.index` deterministically after all reads return. This removes the shared mutable state entirely

### DIFF
--- a/app/src/main/java/org/WenuLink/parameters/ParameterBase.kt
+++ b/app/src/main/java/org/WenuLink/parameters/ParameterBase.kt
@@ -1,6 +1,12 @@
 package org.WenuLink.parameters
 
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.withTimeout
 
 enum class SemanticType {
     BOOL,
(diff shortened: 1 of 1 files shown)


angel-ayala (approved):
Nice solution! That old implementation was only iterating the parameters element and waiting for all the callbacks have a return but definitively this is better.
I was facing something similar, but instead of an NPE I got the `"Duplicate parameter index detected"` message. I will check it later

Issue #65, opened 2026-04-16
Define a versioning and release process for public alpha releases


TimonSchreiber:
For official testing and encouraging third-party contributions, it would be a good idea to not only create tags but also GitHub Releases so other people do not have to build from source prior to testing.

It looks like no release tags currently exist in the repo. I'd suggest we establish a versioning scheme going forward.

We also tested today with the latest version of `develop` and could tag that as a second release. Going forward, I'd suggest aligning on [semver](https://semver.org/) pre-release naming, so the two releases would be `v0.1.0-alpha` and `v0.2.0-alpha` respectively, but happy to hear what others think and how we want to proceed.

angel-ayala:
Silly me, I created a pre-release thinking that with all the architectural changes made, everything was still working but no, so I deleted it.

I did a quick read of the semver specifications, and definitively we are still in `0.x.y` and right now I'm unsure how to defining the subsequent versions of `v0`, since there was not that much planning. So far we are being including features, and implementing lines of codes required in an easy-to-follow architecture.
As we discussed before, I think that we could keep looking to close the version `1.0.0` as our first (pre-)release including everything inside the WenuLink v1 milestone.  So basically, we could still testing the `develop` branch until reaching a less bug stage.

For the Test Procedures we definitively should deliver a version and the compiled APK to correctly follow the aircraft compatibility and its version constraints. That could also encourages new users to have an easy way to test it, I agree. I will look onto major details regards the API key needed and responsibility behind its use.

angel-ayala:
Hi @TimonSchreiber , I been thinking that probably now we can define a `0.x.y` version aiming to close #82 and solve possible bugs related to anything that's been already implemented, unless major changes or another feature be necessary.

TimonSchreiber:
I am all for creating a proper release as soon as possible, but think that the current `ui-rework` should be included since without it, a pre-build release has the hard-coded ip addresses which would be impossible to adjust without touching the source code, defying the purpose of creating a release. I am almost done with finishing out @ElitonSena ui work, only making the changes you and I suggested in our reviews.

If you want I can create the tag and build the apk after #82 is merged for a release draft

angel-ayala:
Sounds good to me, but I still don't know which version should be used. Such definition I think that is conditioned to if we will apply some new features or no. If yes maybe we must set something live `0.5`, and on the contrary `0.9` without the APK. I think that the APK must be available for stable releases only.
What do you think?

[...]

PR #92, opened 2026-04-27
Add release signing config and CI workflow

TimonSchreiber:
Adds the infrastructure needed to produce signed release APKs via CI. Required before cutting `v0.1.0-alpha`.

Related to #65.

### Changes

- `app/build.gradle.kts`: adds a `signingConfigs.release` block that reads from `local.properties`. The release `buildType` applies the config only when a keystore path is configured, so debug and contributor builds without release secrets are unaffected.
- `.github/workflows/release.yml`: new workflow triggered on `v*` tag pushes. Decodes the release keystore from secrets, writes a `local.properties`, builds with `./gradlew assembleRelease`, verifies the APK signature with `apksigner`, and publishes a GitHub Release with the signed APK attached. Tags containing a dash are auto-flagged as pre-releases.

### Required GitHub Actions secrets

The workflow will not succeed until these secrets are populated at the org or repo level:

- `RELEASE_KEYSTORE_BASE64`
- `RELEASE_KEYSTORE_PASSWORD`
- `RELEASE_KEY_ALIAS`
- `RELEASE_KEY_PASSWORD`
- `DJI_API_KEY`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,77 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+permissions:
+  contents: write  # required by softprops/action-gh-release to publish a release
+
+jobs:
+  build-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
+        with:
+          distribution: temurin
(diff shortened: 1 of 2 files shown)


angel-ayala (approved):
Sorry for the late reply @TimonSchreiber , I needed to pay attention to other stuff, I haven't created the email or anything to request the keys required for the action secrets, but don't worry I will do it now and I will ensure to have the keys as Github action secrets.

angel-ayala:
I just added the secrets at repository level, `RELEASE_KEYSTORE_BASE64` needs to be updated, I didn't saw any value for that when created the keystore. Not sure if must to be obtained from Bitwarden or somewhere else.

TimonSchreiber:
See answer [here](https://github.com/WenuLink/wenu-link-android/issues/65#issuecomment-4341865083).

PR #94, opened 2026-04-28
Fix: NaN crash in outgoing telemetry messages

TimonSchreiber:
### Problem

During indoor field testing with a Mavic 2 Pro, the app produced a sustained flood of `IllegalArgumentException: Cannot round NaN value` exceptions from the outgoing-message pipeline. The exceptions are caught downstream so the app keeps running, but every affected MAVLink message is dropped, meaning QGC receives almost no telemetry while the condition persists.

Excerpt from the logcat:

```bash
2026-04-27 16:16:54.xxx ERROR MAVLinkClient: (DefaultDispatcher-worker-11:4040) Error in outgoingMessages java.lang.IllegalArgumentException: Cannot round NaN value.
2026-04-27 16:16:54 ERROR MAVLinkClient: (DefaultDispatcher-worker-13:4042) Error in outgoingMessages java.lang.IllegalArgumentException: Cannot round NaN value.
2026-04-27 16:16:54 ERROR MAVLinkClient: (DefaultDispatcher-worker-3:3914)  Error in outgoingMessages java.lang.IllegalArgumentException: Cannot round NaN value.
2026-04-27 16:16:54 ERROR MAVLinkClient: (DefaultDispatcher-worker-5:3916)  Error in outgoingMessages java.lang.IllegalArgumentException: Cannot round NaN value.
INFO  chatty: uid=10209(org.WenuLink) DefaultDispatch identical 44 lines
ERROR MAVLinkClient: (DefaultDispatcher-worker-3:3914)  Error in outgoingMessages java.lang.IllegalArgumentException: Cannot round NaN value.
INFO  chatty: uid=10209(org.WenuLink) DefaultDispatch identical 71 lines
…
```

### Root cause

DJI's `FlightControllerState` returns NaN for `aircraftLocation.latitude/longitude/altitude` (and related fields) when GPS is unavailable, e.g. indoors. These NaN values flowed unchecked through `FCManager.state2Telemetry` into `TelemetryData`, and from there into MAVLink message builders that scale the values via `kotlin.math.roundToInt()`. Of the rounding helpers, `roundToInt()` is the only one that throws on NaN; `round()` and `Double.toInt()` handle it gracefully. But silently mapping NaN to 0 here would teleport the drone to (0, 0) on QGC's map, which is worse than not sending the message.

### Changes:

- `FCManager.state2Telemetry` sanitizes incoming DJI values via `finiteOr` / `finiteOrNull` extensions. NaN/Inf in lat/lon/altitudes becomes `null`. NaN/Inf in angles and velocities becomes `0`
- `TelemetryData` makes `latitude`, `longitude`, `takeOffAltitude`, `relativeAltitude`, and the derived `altitude` nullable. Other Doubles/Floats remain non-nullable but are now guaranteed finite by the boundary contract
- Message builders that depend on position (`msgGlobalPositionInt`, `msgRawGPSInt`, `msgAltitude`, `msgHUD`) return `null` when required fields are unavailable, which causes the message to be skipped
- `TelemetryMapper.toMavlink` translates `null` -> `Int.MAX_VALUE` (MAVLink's documented unknown dummy value for scaled-int position fields)
- `AircraftHandler.currentCoordinates` aggregates the three nullable fields into `Coordinates3D?`

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
+++ b/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
@@ -33,8 +33,12 @@ class AircraftHandler : CommandHandler<AircraftHandler>() {
     val telemetry = TelemetryHandler.getInstance()
     val currentTelemetry: TelemetryData? get() = telemetry.getData()
     val currentCoordinates: Coordinates3D?
-        get() = telemetry.getData()?.let {
-            Coordinates3D(it.latitude, it.longitude, it.relativeAltitude)
+        get() {
+            val data = telemetry.getData() ?: return null
+            val lat = data.latitude ?: return null
+            val lon = data.longitude ?: return null
+            val rel = data.relativeAltitude ?: return null
+            return Coordinates3D(lat, lon, rel)
         }
     var isPowerOff = true
     val parameters by lazy {
--- a/app/src/main/java/org/WenuLink/adapters/aircraft/TelemetryData.kt
+++ b/app/src/main/java/org/WenuLink/adapters/aircraft/TelemetryData.kt
@@ -15,8 +15,8 @@ data class TelemetryData(
     val roll: Double,
     val pitch: Double,
     val yaw: Double,
-    val latitude: Double,
-    val longitude: Double,
+    val latitude: Double?,
+    val longitude: Double?,
     val positionX: Float,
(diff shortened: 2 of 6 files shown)


angel-ayala (approved):
Well done! I was facing some issues related to this without GPS information as well. Thanks for the patch.

PR #95, opened 2026-04-28
Split address configuration into separate IP and port fields

TimonSchreiber:
- Store IP (String) and port (Int) as separate keys in WenuLinkPreferences
- Update SettingsViewModel to expose ip/port as separate StateFlows with independent setters
- Change updateGCSAddress and updateServerAddress to take (ip, port) directly, removing the spliting
- Rewrite AddressScreen with two OutlinedTextFields: IP uses KeyboardType.Uri to allow hostnames, port uses KeyboardType.Number capped at five digits. Tap-to-edit replaces the explicit edit-mode toggle, with isServiceRunning as the sole edit guard. Each field commits independently on focus loss when valid; the keyboard's Done action clears focus to trigger the same path.
- Drop the copy-to-clipboard control. long-press text selection covers the rare case.
- Validation is lenient on the IP side (reject empty, ':', '/', and whitespace) and strict on the port (1..65535).

### DIFF
--- a/app/src/main/java/org/WenuLink/WenuLinkPreferences.kt
+++ b/app/src/main/java/org/WenuLink/WenuLinkPreferences.kt
@@ -6,14 +6,22 @@ import androidx.core.content.edit
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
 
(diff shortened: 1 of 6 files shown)


angel-ayala (approved):
Great improvement with clean changes, I didn't see anything out of place. Amazing job!

PR #96, opened 2026-04-28
Feature: BridgeHealth on Main Screen

TimonSchreiber:
Adds a lightweight bridge-health readout to the main screen, replacing the `telemetrySummary` slot previously removed in #82. Surfaces the diagnostic signals that QGC and the RC display can't show: whether the WenuLink bridge itself is alive and producing output, for use during initial bring-up (bad IP, firewall blocking the port, no GCS connection yet).

### DIFF
--- a/app/src/main/java/org/WenuLink/mavlink/MAVLinkClient.kt
+++ b/app/src/main/java/org/WenuLink/mavlink/MAVLinkClient.kt
@@ -10,6 +10,7 @@ import java.net.InetAddress
 import java.net.SocketException
 import java.net.SocketTimeoutException
 import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicLong
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.delay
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/WenuLink/mavlink/MAVLinkClient.kt
```diff
@@ -26,6 +27,14 @@ class MAVLinkClient(
...
     val mustReceiveMessages = AtomicBoolean(false)
     val mustSendMessages = AtomicBoolean(false)
+    val rxCount = AtomicLong(0L)
+    val txCount = AtomicLong(0L)
+    val parseErrors = AtomicLong(0L)
+
+    @Volatile
+    var lastObservedPeer: String? = null
+        private set
+
```
angel-ayala: Apparently this attribute is not being used, probably from draft

angel-ayala (dismissed):
Nice addition @TimonSchreiber ! , definitively this will help to monitor statistics from messages on the app. I haven't tested yet, but everything looks good to me. Only one detail with a variable that apparently is not being used.

app/src/main/java/org/WenuLink/mavlink/MAVLinkClient.kt
TimonSchreiber: Sorry about that!
I had initially planned to add an attribute to check if the configured mavlink address and the one sending the heartbeats differ, but couldn't really think of a situation where that would actually happen. The only situation I was able to think of was when two QGC instances were running on the same network.
So I went and removed that feature again, but must have overlooked this attribute in `MAVLinkClient`.

angel-ayala: No worries, that kind of changes I think need to be addressed in a future if needed, I think for example in a multi-agent situation requiring different aircraft to share information between each other. Seems to be a good feature for next release maybe.

angel-ayala (approved):
I didn't review it again, just approving the rebase

PR #99, opened 2026-04-29
Release v0.1.0-alpha

TimonSchreiber:
First official pre-release version v0.1.0-alpha


angel-ayala (approved):
Finally we achieve this point! I'm aware of all changes and proud to accept this! Let's check if the workflow produces the APK without any error

PR #100, opened 2026-04-29
Main

TimonSchreiber:
Or would you prefer just merging the version bump branch into `develop` instead of `main` @angel-ayala?


angel-ayala (approved):
I cannot think in any problem if is this way or the other. So this is ok.

PR #111, opened 2026-06-04
Fix BATTERY_STATUS message

### LINKED ISSUE #106: SYS_STATUS pack crashes with IllegalArgumentException: -1 outside unsigned short range
**Environment**
 
- WenuLink: v0.1.1-alpha
- Drone: DJI Mavic Pro
- Android device: Lenovo TB-X606X (Android 10)
- GCS: QGroundControl v5.0.8
 
**What happened**
 
During a mission test session the app crashed. Android auto-restarted `WenuLinkService` shortly after, and the MAVLink service toggle in the UI was off. The drone was not airborne at the time of the crash.
 
```
FATAL EXCEPTION: DefaultDispatcher-worker-8
Process: org.WenuLink, PID: 25329
java.lang.IllegalArgumentException: Value is outside of the range of an unsigned short: -1
	at com.MAVLink.Messages.MAVLinkPayload.putUnsignedShort(MAVLinkPayload.java:162)
	at com.MAVLink.common.msg_sys_status.pack(msg_sys_status.java:153)
	at org.WenuLink.mavlink.MAVLinkClient$sendMessage$2.invokeSuspend(MAVLinkClient.kt:102)
```
 
**Additional context**
 
QGC displayed battery 0% throughout the session, suggesting battery telemetry into SYS_STATUS is broken upstream of the pack call.
 
**Logs**
 
See parent compatibility report (#105), logcat from the 14:48:18 export.

angel-ayala:
Updated the voltage cell information with a fixed length array with the total battery voltage as first element

Partially fixes #106

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/aircraft/TelemetryData.kt
+++ b/app/src/main/java/org/WenuLink/adapters/aircraft/TelemetryData.kt
@@ -174,10 +174,15 @@ data class MAVLinkBatteryData(
 object BatteryMapper {
     private const val UNKNOWN_INT = -1
     private const val UNKNOWN_SHORT = Short.MAX_VALUE // INT16_MAX per MAVLink spec
+    private const val UINT16_MAX = 65535
 
     fun toMavlink(source: BatteryData): MAVLinkBatteryData {
         val fullCharge = source.fullChargeCapacity
         val remaining = source.chargeRemaining
+        val voltages = mutableListOf(source.voltage ?: UINT16_MAX).apply {
+            while (size < 10) add(UINT16_MAX)
+        }.toIntArray()
+
         return MAVLinkBatteryData(
             currentConsumed = if (fullCharge != null && remaining != null) {
                 fullCharge - remaining
@@ -187,7 +192,7 @@ object BatteryMapper {
             temperature = source.temperature
                 ?.let { (it * 100).toInt().toShort() }
                 ?: UNKNOWN_SHORT,
-            voltages = source.voltageCells ?: emptyList(),
+            voltages = voltages.toList(),
             currentBattery = source.current
                 ?.let { (it * 10).toShort() }
                 ?: UNKNOWN_INT.toShort(),
--- a/app/src/main/java/org/WenuLink/adapters/aircraft/TelemetryData.kt
+++ b/app/src/main/java/org/WenuLink/adapters/aircraft/TelemetryData.kt
@@ -174,14 +174,18 @@ data class MAVLinkBatteryData(
 object BatteryMapper {
     private const val UNKNOWN_INT = -1
     private const val UNKNOWN_SHORT = Short.MAX_VALUE // INT16_MAX per MAVLink spec
-    private const val UINT16_MAX = 65535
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/WenuLink/adapters/aircraft/TelemetryData.kt
```diff
@@ -174,10 +174,15 @@ data class MAVLinkBatteryData(
 object BatteryMapper {
     private const val UNKNOWN_INT = -1
     private const val UNKNOWN_SHORT = Short.MAX_VALUE // INT16_MAX per MAVLink spec
+    private const val UINT16_MAX = 65535
```
TimonSchreiber:
```suggestion
    private const val UINT16_MAX = UShort.MAX_VALUE
```

app/src/main/java/org/WenuLink/adapters/aircraft/TelemetryData.kt
```diff
@@ -174,10 +174,15 @@ data class MAVLinkBatteryData(
...
 
     fun toMavlink(source: BatteryData): MAVLinkBatteryData {
         val fullCharge = source.fullChargeCapacity
         val remaining = source.chargeRemaining
+        val voltages = mutableListOf(source.voltage ?: UINT16_MAX).apply {
+            while (size < 10) add(UINT16_MAX)
+        }.toIntArray()
+
```
TimonSchreiber:
```suggestion
        val voltages = IntArray(10) {
            if (it == 0) {
                source.voltage ?: UINT16_MAX.toInt()
            } else {
                UINT16_MAX.toInt()
            }
        }

```

app/src/main/java/org/WenuLink/adapters/aircraft/TelemetryData.kt
TimonSchreiber:
```suggestion
            voltagesBattery = source.voltage ?: UINT16_MAX.toInt(),
```

TimonSchreiber (changes requested):
Hi @angel-ayala ,

I am sorry I delayed this review so long.

Two minor suggestions on readability:
1. Use Kotlin's build-in `UShort.MAX_VALUE` instead of a hardcoded 65355, most programmers might recognize it, but this way the intent is much clearer.
2. The construction of a mutable list and the conversion back to an IntArray is redundant, I added a suggestion which creates the `IntArray` directly, but there are many ways to write this in a more straight forward way.

Regarding this partially fixing of #106 :
I think the underlying issue there is not the voltages list, althought this is a nice improvement, but the `voltagesBattery` field, which has the wrong default value, see MavLink documentation or the auto-generated MavlInk class for `msg_sys_status.java`:
```java
/**
 * Battery voltage, UINT16_MAX: Voltage not sent by autopilot
 */
@Description("Battery voltage, UINT16_MAX: Voltage not sent by autopilot")
@Units("mV")
public int voltage_battery;   
```

angel-ayala:
Good catch @TimonSchreiber , I just updated the code and applied your suggestions.

TimonSchreiber (approved):
Looks good to me, ready to merge!

PR #112, opened 2026-06-15
Fix/mission upload reset

### LINKED ISSUE #109: Waypoint count not reset between consecutive mission uploads
**Environment**
 
- WenuLink: v0.1.1-alpha
- Drone: DJI Mavic Pro
- GCS: QGroundControl v5.0.8
 
**What happened**
 
A second mission upload after a successful first upload, or after a direct takeoff with no prior mission, fails with a waypoint count mismatch. Force-stopping and restarting the app clears the state.
 
**Reproductions observed**
 
- After a successful mission upload, uploading a different mission fails with mismatched waypoint count.
- After a direct takeoff (no mission), a subsequent mission upload fails with mismatched waypoint count.
 
**Logs**
 
angel-ayala:
Main changes were made in `MissionCommands` and `RequestCommands`  with the temporal mission list handled by `MissionAssembler` being used during the MAVLink plan flight upload process. 
The mission is then uploaded to the aircraft after checking the home position and before start the mission in the `RequestStartMission` command.
Since the upload process sometimes fails, a maximum of 3 automatic retry was defined before command's fail.

Other code reorganization regards classes concern was also included.

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/mission/MissionCommands.kt
+++ b/app/src/main/java/org/WenuLink/adapters/mission/MissionCommands.kt
@@ -54,12 +54,14 @@ data class UploadWaypointMission(val maxTries: Int = 3) : MissionCommand {
         }
     }
     override suspend fun execute(ctx: MissionHandler): UnitResult {
-        repeat(maxTries) {
-            val uploadResult = tryUpload(ctx)
-            if (!uploadResult.hasError) return UnitResult.ok
+        val maxAttempts = maxTries.coerceAtLeast(1)
+        var attempt = 0
+        var result: UnitResult = CommandResult.error("No upload attempted")
+        while (result.hasError && attempt < maxAttempts) {
+            result = tryUpload(ctx)
+            attempt++
         }
-        // If all retries failed, do final attempt and return result
-        return tryUpload(ctx)
+        return result
     }
 
     override suspend fun onStop(ctx: MissionHandler) {
--- a/app/src/main/java/org/WenuLink/adapters/mission/MissionCommands.kt
+++ b/app/src/main/java/org/WenuLink/adapters/mission/MissionCommands.kt
@@ -35,7 +35,7 @@ data class UploadMissionCommand(
     private val flightSpeed: Float = 5f
 ) : MissionCommand {
     override fun validate(ctx: MissionHandler): UnitResult = when {
-        ctx.state.canCreateMission() -> CommandResult.ok
+        ctx.state.canUploadMission -> CommandResult.ok
         else -> CommandResult.error("Upload not ready")
     }
 
@@ -55,9 +55,9 @@ data class UploadMissionCommand(
 
 data object StartWaypointMission : MissionCommand {
     override fun validate(ctx: MissionHandler): UnitResult = when {
-        ctx.state.canCreateMission() -> CommandResult.error("No mission found")
-        ctx.state.isActive() -> CommandResult.error("Already started")
-        ctx.state.canStartMission() -> CommandResult.ok
+        ctx.state.canUploadMission -> CommandResult.error("No mission found")
+        ctx.state.isActive -> CommandResult.error("Already started")
+        ctx.state.canStartMission -> CommandResult.ok
         else -> CommandResult.error("Not ready")
     }
 
@@ -77,7 +77,7 @@ data object StartWaypointMission : MissionCommand {
 
 data object PauseWaypointMission : MissionCommand {
     override fun validate(ctx: MissionHandler): UnitResult = when {
-        ctx.state.isActive() -> CommandResult.ok
+        ctx.state.isActive -> CommandResult.ok
         else -> CommandResult.error("Not started")
     }
 
@@ -97,7 +97,7 @@ data object PauseWaypointMission : MissionCommand {
 
 data object ResumeWaypointMission : MissionCommand {
     override fun validate(ctx: MissionHandler): UnitResult = when {
-        ctx.state.isPaused() -> CommandResult.ok
+        ctx.state.isPaused -> CommandResult.ok
         else -> CommandResult.error("Already in execution")
     }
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/WenuLink/adapters/mission/MissionCommands.kt
```diff
@@ -30,107 +30,139 @@ sealed interface MissionCommand : ICommand<MissionHandler> {
...
         }
+    }
+    override suspend fun execute(ctx: MissionHandler): UnitResult {
+        repeat(maxTries) {
+            val uploadResult = tryUpload(ctx)
+            if (!uploadResult.hasError) return UnitResult.ok
+        }
+        // If all retries failed, do final attempt and return result
+        return tryUpload(ctx)
+    }
```
TimonSchreiber, line 65:
```suggestion
    }

    override suspend fun execute(ctx: MissionHandler): UnitResult {
        val maxAttempts = maxTries.coerceAtLeast(1)
        var attempt = 0
        var result: UnitResult = CommandResult.error("No upload attempted")
        while (result.hasError) {
            result = tryUpload(ctx)
            attempt++
            if (attempt >= maxAttempts) break
        }
        return result
    }
```

app/src/main/java/org/WenuLink/adapters/mission/MissionHandler.kt
```diff
@@ -63,16 +78,17 @@ data class MissionState(
...
     fun reset(): MissionState = copy(
         startSequence = 1,
         currentSequence = null,
         isComplete = false,
         unvisitedSequence = true
     ).fromMissionManager()
 
-    fun clearWaypoints() = assembler.reset()
+    fun resetAssembler(): MissionState {
+        assembler?.reset()
```
TimonSchreiber:
```suggestion
```

app/src/main/java/org/WenuLink/adapters/mission/MissionHandler.kt
```diff
@@ -143,117 +161,100 @@ class MissionHandler : CommandHandler<MissionHandler>() {
...
 
-            else -> return false
+        if (state.hasAssembler) {
+            logger.d { "Clearing previous mission assembler" }
+            state = state.resetAssembler()
+            state = state.setUploadReady(false)
         }
-        return true
+        logger.d { "Adding new mission id 202606" }
+        state = state.createMission(202606)
```
TimonSchreiber: Hardcoding "June 2026" here?

app/src/main/java/org/WenuLink/adapters/mission/MissionHandler.kt
```diff
@@ -25,22 +18,37 @@ data class MissionState(
     val id: Int = 202512,
```
TimonSchreiber: Hardcoded to "December 2025"

app/src/main/java/org/WenuLink/mavlink/controllers/NavigationController.kt
```diff
@@ -261,8 +261,8 @@ class NavigationController(
 
     fun sendStatusText(status: String, severity: Int) = client.sendMessage(
         msg_statustext().apply {
-            logger.d { "sendStatusText" }
-            text = status.toByteArray()
+            logger.d { "sendStatusText $status" }
+            text = status.chunked(50).first().toByteArray()
```
TimonSchreiber:
```suggestion
            text = status.take(50).toByteArray()
```

TimonSchreiber (changes requested):
Great PR @angel-ayala going way beyond #109 !

I have no major complaints, just a few minor suggestions (see inline comments):

1. In the `MissionCommands.execute` method you use the `repeat` function, which seems nice at first, but suggests that the intended flow is to actually call the lambda `maxTries` times, but the ideal run would be to only call it once. I think a while-loop fits better because the break condition is when `tryUploads` retuns `CommandResult.ok` and we loop until this happens, only adding a safeguard limit to the number of iterations secondarily. In my opinion this should be a while-loop. This also makes the total number of iterations clear, the current version attempts up to 3 times and then once more.
2. Reseting the assembler befor throwing it away and returning a copy of `MissionState` without it is not necessary.
3. There are two hardcoded ids in MIssionCommands which seem to represent dates, can we find better values here?
4. chunking a string and then taking the first element of that new array can be simplyfied to just taking the first 50 chars of a string.

Everything else looks really good to me.

app/src/main/java/org/WenuLink/adapters/mission/MissionHandler.kt
```diff
@@ -143,117 +161,100 @@ class MissionHandler : CommandHandler<MissionHandler>() {
...
-
-        state.assembler.addWaypoint(coordinates)
-
-        // Delay (seconds)
-        if (params.holdTimeSec > 0f) {
-            state.assembler.addActionToLast(DelayAction((params.holdTimeSec * 1000).toLong()))
-        }
+    fun processItem(itemMsg: msg_mission_item_int): Boolean {
+        if (!state.hasAssembler) logger.w { "Processing item without assembler" }
+        return state.assembler?.addWaypointNode(itemMsg) ?: false
```
TimonSchreiber:
```suggestion
        if (!state.hasAssembler) {
            logger.w { "processItem called without assembler, item ${itemMsg.command} dropped"
            return false
        }
        return state.assembler?.addWaypointNode(itemMsg) ?: false
```

TimonSchreiber (changes requested):
I think returning false explicitly in case of a missing assembler makes the intent more clear.

angel-ayala:
I see your point regards to 1. I just changed to avoid the use of mutable variables, but definitively a while-loop fits better.
Regards 3. such value was only an arbitrary int value to define the mission ID, which right now only a single value is used since we does not have multiple mission management.
I just updated the suggestions of point 2 and 4, with the last need to be improved in another issue/PR since is regarding to MAVLink logs.

TimonSchreiber (approved):
Also ready to merge.

PR #113, opened 2026-06-22
Fix/async state

### LINKED ISSUE #108: Mission start sends arm to already armed/flying aircraft and fails
**Environment**
 
- WenuLink: v0.1.1-alpha
- Drone: DJI Mavic Pro
- GCS: QGroundControl v5.0.8
 
**What happened**
 
When the drone is already airborne (manual takeoff via RC, or already flying after a previous mission start), starting a mission from QGC fails with "failed to arm". WenuLink does not appear to check whether the aircraft is already armed before issuing the arm command.
 
**Reproduction**
 
1. Take off manually with the RC.
2. From QGC, attempt to start a mission.
3. Mission start fails with "failed to arm".
 
**Logs**
 
### LINKED ISSUE #110: Flight-mode state in WenuLink does not track actual aircraft state
**Environment**
 
- WenuLink: v0.1.1-alpha
- Drone: DJI Mavic Pro
- GCS: QGroundControl v5.0.8
 
**What happened**
 
WenuLink's flight-mode state stays out of sync with the aircraft in three observed cases:
 
1. RC manual takeoff: state remains "Ready to Fly" rather than transitioning to "Flying".
2. RC force-land during a mission: state remains "Flying" rather than transitioning to "Landed".
3. After an RC-initiated landing: state is stuck at "Landing".
 
In all three cases force-stopping and restarting the app is the only known way to recover the state.
 
**Logs**
 
angel-ayala:
The presented changes aims to enforce the concerns of `AircraftStateMachine` to properly sync the current `AircraftState` through iterative `AircraftHandler.syncState` calls on the `monitorJob`, aiming to solve #110  and #108.

Current tests with P4P ensured the proper sync on arm and flying state reflected in QGC. Nevertheless, arm triggers and mission start operations from QGC still need to be check.

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
+++ b/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
@@ -14,27 +14,27 @@ sealed interface AircraftCommand : ICommand<AircraftHandler> {
 data class BootCommand(val timeout: Long = 5000L) : AircraftCommand {
     override fun validate(ctx: AircraftHandler): UnitResult = when {
         !ctx.isPowerOff -> CommandResult.error("Already booted")
-        else -> ctx.canDispatchTransition(BootTransition)
+        else -> ctx.stateMachine.canDispatch(BootTransition)
     }
 
     override suspend fun execute(ctx: AircraftHandler): UnitResult {
         // TODO: check if compatible with CancellableCoroutine
-        ctx.dispatchTransition(BootTransition)
+        ctx.stateMachine.dispatch(BootTransition)
         val bootResult = ctx.boot(timeout)
         if (bootResult.hasError) return bootResult
-        ctx.dispatchTransition(StandbyTransition)
+        ctx.stateMachine.dispatch(StandbyTransition)
         return CommandResult.ok
     }
 
     override suspend fun onStop(ctx: AircraftHandler) {
-        ctx.dispatchTransition(InitialTransition)
+        ctx.stateMachine.dispatch(InitialTransition)
     }
 }
 
 data class ArmCommand(val timeout: Long = 5000L) : AircraftCommand {
     override fun validate(ctx: AircraftHandler): UnitResult = when {
         !ctx.sensorsHealthy -> CommandResult.error("Sensors failing")
-        else -> ctx.canDispatchTransition(ArmTransition)
+        else -> ctx.stateMachine.canDispatch(ArmTransition)
     }
 
     override suspend fun execute(ctx: AircraftHandler): UnitResult {
@@ -47,7 +47,7 @@ data class ArmCommand(val timeout: Long = 5000L) : AircraftCommand {
             }
         }
         // Automatic takeoff only. Must wait for state changes
-        ctx.dispatchTransition(ArmTransition)
+        ctx.stateMachine.dispatch(ArmTransition)
 
         return CommandResult.ok
     }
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
```diff
@@ -257,6 +222,11 @@ class AircraftHandler : CommandHandler<AircraftHandler>() {
     }
 
     fun takeOff() {
+        // sync and check for already-airborne cases
+        if (state.isFlying()) {
+            logger.d { "Aircraft already airborne" }
```
```suggestion
            logger.w { "takeOff() called but aircraft already airborne, skipping" }
```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftHandler.kt
```suggestion
if (error != null) logger.e { "Takeoff error: $error" }
```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftState.kt
```diff
@@ -248,6 +239,24 @@ class AircraftStateMachine {
...
+            dispatch(TakeoffTransition)
+        }
+        // is flying
+        if (fcState.isFlying() && state.isTakingOff()) {
+            dispatch(FlyingTransition)
+        }
+        // is landing should be triggered and fixed until standby
+        if (!fcState.isArmed() && !fcState.isFlying() && !state.armRequested) {
+            dispatch(StandbyTransition)
+        }
```
TimonSchreiber:
```suggestion
    when {
        // RC armed while on ground: advance to armed
        fcState.isArmed() && state.isOnTheGround() -> dispatch(ArmTransition)
        // Armed and on ground: advance to takeoff
        fcState.isArmed() && fcState.isFlying() && state.isOnTheGround() ->
            dispatch(TakeoffTransition)
        // Taking off and now flying: advance to flying
        fcState.isArmed() && fcState.isFlying() && state.isTakingOff() ->
            dispatch(FlyingTransition)
        // Disarmed and grounded: return to standby
        !fcState.isArmed() && !fcState.isFlying() && !state.armRequested ->
            dispatch(StandbyTransition)
    }
```

app/src/main/java/org/WenuLink/adapters/aircraft/AircraftCommands.kt
```diff
@@ -47,7 +47,7 @@ data class ArmCommand(val timeout: Long = 5000L) : AircraftCommand {
             }
         }
         // Automatic takeoff only. Must wait for state changes
-        ctx.dispatchTransition(ArmTransition)
+        ctx.stateMachine.dispatch(ArmTransition)
```
TimonSchreiber, line 51:
```suggestion

    val armed = ctx.waitArmTransition(true, timeout)
    if (!armed) {
        ctx.stateMachine.dispatch(StandbyTransition)
        return CommandResult.error("Arm request timed out")
    }

```

TimonSchreiber (changes requested):
Great work as always @angel-ayala 

Here are some thought I had:

1. I am not sure that removing the delegation methods `canDispatchTransition` and `dispatchTransition` from `AircraftHandler`, but instead expose `stateMachine` as a public property and make `AircraftCommands` use it directly is necessarily an improvement, but if you prefer it like this, that is fine with me.
2. `AircraftHandler.takeOff`: Elevating the new log statement in `takeOff` from debug to warning and using logger.e for the `startTakeoff` error.
3. `AircraftState.sync`: Doing a double dispatch call in one of the branches is inconsistent with the other cases and separating them only adds a minor delay befor the state catches up with the real aircraft (one monitorJob-loop). Also using a when-expression makes it clear that only one of the branches can be executed.
4. `ArmCommand.execute`: In case the arm fails after the `ArmTransition` is dispatched, `sync()` can not recover the state to standby. My suggestion: adding a timeout after the dispatch call.

angel-ayala:
Hi @TimonSchreiber , sorry for the late reply I was busy with other things last weeks.

Regards to your observation:
1. I just exposed the `stateMachine` and therefore the `dispatch` and `canDispatch` methods since improve the traceability on where its being used, simplifying the development of this PR. Such methods were only being mirrored from the `AircraftHandler` class with no additional logic. So considering that commands are tight up with such handler, I think there is no major issue in leaving this way.
2. I totally agree, done
3. Very nice implementation, I didn't thought on that
4. I'm unsure about this change, since such command is called to start the mission, only must set the arm flag until the vehicle be armed. Instead, if the arm condition does not succeed, the `sync` method called from the `monitorJob` now should catch such condition. I changed the attribute as `armTimestamp` to compare how long is taking to arm, with a maximum value of 10 seconds.

TimonSchreiber (approved):
Your solution is better than what I had suggested, ready to merge!

PR #114, opened 2026-07-03
Fix/mission item altitude

### LINKED ISSUE #107: Drone climbs past requested takeoff altitude (Mavic Pro)
**Environment**
 
- WenuLink: v0.1.1-alpha
- Drone: DJI Mavic Pro
- GCS: QGroundControl v5.0.8
 
**Mission**
 
5-item plan (NAV_TAKEOFF, 2x NAV_WAYPOINT, NAV_DELAY, NAV_RETURN_TO_LAUNCH), spatial items in MAV_FRAME_GLOBAL_RELATIVE_ALT. See `.plan` attached to parent #105.
 
**What happened**
 
Two attempts with different takeoff altitudes:
 
1. Takeoff altitude 50 m. The drone continued climbing past 50 m after mission start.
2. Takeoff altitude reduced to 20 m. The drone again continued climbing past 20 m.
 
In both cases the drone was stopped manually via RC at roughly 55 m, with no indication that the climb would have stopped on its own.
 
**Logs**
 
TimonSchreiber:
Fixes #107 

#### Changes

- New MissionNode.Home stores the seq 0 home item as metadata at node index 0. It is excluded from the DJI mission and re-emitted in MAV_FRAME_GLOBAL on download. Node index continues to equal MAVLink seq.
- Item assembly returns a sealed ItemAssemblyResult instead of a Boolean, enabling MAV_MISSION_UNSUPPORTED_FRAME NACKs. Only MAV_FRAME_GLOBAL_RELATIVE_ALT(_INT) is accepted for coordinate items; AMSL is rejected rather than converted, since MSDK v4 provides no reliable home AMSL reference.
- NAV_TAKEOFF is materialized as the first DJI waypoint, so the initial climb targets the planned takeoff altitude. Takeoff items with (0, 0) coordinates are rejected at upload. The waypoint count is derived from the builder's waypoint list.

Branched off #112 and intended to be merged after it, since both touch the mission upload path. Will rebase onto develop once #112 lands.

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/mission/ItemAssemblyResult.kt
+++ b/app/src/main/java/org/WenuLink/adapters/mission/ItemAssemblyResult.kt
@@ -1,0 +1,8 @@
+package org.WenuLink.adapters.mission
+
+sealed interface ItemAssemblyResult {
+    data object Accepted : ItemAssemblyResult
+    data object UnsupportedCommand : ItemAssemblyResult
+    data object UnsupportedFrame : ItemAssemblyResult
+    data object NoActiveMission : ItemAssemblyResult
+}
--- a/app/src/main/java/org/WenuLink/adapters/mission/MissionAssembler.kt
+++ b/app/src/main/java/org/WenuLink/adapters/mission/MissionAssembler.kt
@@ -2,6 +2,8 @@ package org.WenuLink.adapters.mission
 
 import com.MAVLink.common.msg_mission_item_int
 import com.MAVLink.enums.MAV_CMD
+import com.MAVLink.enums.MAV_FRAME
+import io.getstream.log.taggedLogger
 import org.WenuLink.adapters.aircraft.Coordinates3D
 import org.WenuLink.mavlink.messages.ConditionYawMessage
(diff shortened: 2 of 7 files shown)


angel-ayala (commented):
Nice work @TimonSchreiber , the only comment that I have is regards the title. The addressed changes make sense to me, but were focused on mission items altitude instead the aircraft take off. So I will update the PR title.

I agree that #112 must be merged first.

What worries me the most is if such changes were already tested with the aircraft.

I was also aimed to address the problem with the altitude during take off, focused on the take off action that can be triggered from QGC, but I haven't tested it yet. I will open a PR of such changes to check and compare in case of requiring to update other lines.

TimonSchreiber:
Definitely a better name for this pr.

Do you mean, you are worried to approve this without tests or what the latest test flights say about this problem?

angel-ayala:
Yes, my worry is about approve the PR without testing. But if all seems to be ok, maybe a new PR can be opened with a better fix.

angel-ayala (approved):
Nice work, everything seems to be ok. Ready to merge.

PR #115, opened 2026-07-03
Minor refactor

TimonSchreiber:
Android Studio flagged this as an error in future versions:

```
Return in function with expression body and without explicit return type. Use block body '{...}' or add an explicit return type. This will become an error in language version 2.5.
```

Doesn't change much but prevents future problems.


angel-ayala (approved):
I see no problem with such changes, LGTM.

PR #116, opened 2026-07-11
Add relative altitude on GoToAction

angel-ayala:
Introduces a POC to set the takeoff altitude relative to the current aircrat location as demanded in the SDK documentation for the GoToAction

Related to #107

The public logger of `WenuLinkHandler` is only to check if everything is set

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/RequestCommand.kt
+++ b/app/src/main/java/org/WenuLink/adapters/RequestCommand.kt
@@ -86,12 +86,19 @@ data class RequestTakeoff(val altitude: Float = 2f, val timeout: Long = 15_000L)
         val takeoffResult = ctx.dispatchAndAwait(WenuLinkCommand.Aircraft(TakeoffCommand(timeout)))
         if (takeoffResult.hasError) return takeoffResult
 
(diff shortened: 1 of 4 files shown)


TimonSchreiber:
I am not sure about this pr @angel-ayala.

Based on the dji documentation I see no reason to believe that the altitude in the `GoToAction` is a relative altitude, see [dji-api](https://developer.dji.com/api-reference/android-api/Components/Missions/DJIGoToAction.html#djigotoaction_initwithcoordinatewithaltitude_inline)
The phrase "from the current aircraft position" describes the start of the movement, not the reference frame of the altitude. It is not ideally worded, but otherwise the paramater would be named "relativeAltitude" or something similar.

Testing this in a simulator might help, but I am not convinced this solves the problem.

angel-ayala:
I understand what you mean.
I just tested with the simulation feature and works the same way that the version without this relative altitude, so it must be tested with the real aircraft to see the actual behavior.

mjohenneken:
I'm gonna check if i can test this in the mavic 2 assistant emulator. Not sure if i can set the starting to position != 0 altitude which would be required for testing.

Else I have to test this outside

PR #118, opened 2026-07-30
Fix conditions for proper state synchronization

### LINKED ISSUE #117: [Bug]: State synchronization P4P
### WenuLink App Version

5d67f08

### Drone Model

Phantom 4 Pro

### Test Environment

Real drone

### Android Device (running WenuLink)

Redmi Note 10S

### Android Version

13

### Remote Controller (if separate)

_No response_

### GCS Software & Version

QGC

### Related MAVLink Microservice

- [ ] Heartbeat/Connection
- [ ] Mission
- [ ] Parameter
- [ ] Command
- [ ] Camera
- [x] Other (describe below)

### Related MAVLink Message(s)

_No response_

### Steps to Reproduce

1. Launch App
2. Connec Drone
3. Start services
4. Manually arm vehicle from RC
5. Manually disarm vehicle from RC

### Reproducibility

Always

### Expected Behavior

Proper MAVLink Armed and Ready to flight state synchronization

### Actual Behavior

Reported MAVLink state in QGC loops between Armed and Flying

### Logs

_No response_

### Additional Notes

Current code version also present the bug calling the takeoff button from QGC, I just applied a local path, but still need to be improved and tested

angel-ayala:
Improves the conditions for the asynchronous state transitions in `AicraftStateMachine.sync()`

Fix #117

### DIFF
--- a/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftState.kt
+++ b/app/src/main/java/org/WenuLink/adapters/aircraft/AircraftState.kt
@@ -247,7 +247,8 @@ class AircraftStateMachine {
         val fcState = state.resolveFrom(isArmed, isFlying)
         when {
             // RC trigger arm while on ground: advance to armed
-            fcState.isArmed() && state.isOnTheGround() -> dispatch(ArmTransition)
+            fcState.isArmed() && !state.isArmed() && state.isOnTheGround() ->
+                dispatch(ArmTransition)
 
             // Armed and on ground: advance to takeoff
             fcState.isArmed() && fcState.isFlying() && state.isOnTheGround() ->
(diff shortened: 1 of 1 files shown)


mjohenneken:
do you expect this to happen on other models ?
Shall i help in testing this?

angel-ayala:
I was looking a little bit into the code, and I discovered a two major things:
1. The state transitions are being dispatched from the commands and the `sync()` method in a sequential and poll ways, respectively.
2. The `AircraftState` definitions mixes the armRequested flag for handling ARM-TAKEOFF message sequences from QGC and only call the takeoff method from the SDK and the `isArmed()` status which requires to be redefined.

Therefore, there will be a group of state transitions that will be handled in a sequential manner, directly from the commands, and another group related to the aircraft arm and flying changes  handled in a polling fashion. Additionally, it is required to split the concerns into state synchronization and arm-takeoff sequence flags.

My last changes corrected the state transition conditions from the `sync()` method, tested manually by arming and disarming with the RC, but now is unable to take off from the QGC UI.
I will look up how to address such changes and I will update the branch soon.

angel-ayala:
I think now should works, I will test it and let you know. If you can test it should be nice too.

angel-ayala:
Now everything seems to works fine, I just did some tests today and corrected a few things.
Major changes were related to handle the arm, takeoff, flying, and standby states entirely in the `sync()` as async fashion, with boot and shutdown transitions handled in a sequential manner.
Therefore, all dispatch calls outside `sync()` must be avoided. The single exception was the `LandTransition`, since it is manually requested to switch the flight mode and begin to descend, the manual transition allows to asynchronously wait for the aircraft touch ground.

Nevertheless, as this is one of the main changes addressing an asynchronous behavior, I noticed that most of methods were defined focused on a sequential fashion, requiring to update such definitions into an async fahion.
However, I think that such change should be discussed in another issue or PR.
There are situations. as the arm timeout condition, that require flag value changes (`mustArm`) from the command side to later handle the arm request called on the polling `sync()` method.
The latter is just an ideia considering the current arquitecture, I think that should be better async fashion approaches to re-define `sync()` in the future.

PR #123, opened 2026-09-22
Fix mission upload takeoff node

angel-ayala:
Some last update introduced a bug since takeoff mission item only have altitude parameter.
When the takeoff has no coordinates information, will be replaced by the home node coordinates.

All tests passed on simulation feature.


app/src/main/java/org/WenuLink/adapters/mission/MissionAssembler.kt
```diff
@@ -118,10 +116,16 @@ class MissionAssembler(private val id: Int) {
 
     private fun assembleTakeoffNode(itemMsg: msg_mission_item_int): ItemAssemblyResult {
         if (!hasSupportedFrame(itemMsg)) return ItemAssemblyResult.UnsupportedFrame
+
         val params = NavTakeoffMissionItem(itemMsg)
-        addTakeoff(
-            Coordinates3D(params.latitude, params.longitude, params.altitude)
-        )
+        // If there is no coordinates on takeoff node, replace with the first node
+        // assumed as Home given seq == 0 f
```
TimonSchreiber:
```suggestion
        // assumed as Home given seq == 0
```

app/src/main/java/org/WenuLink/adapters/mission/MissionAssembler.kt
```diff
@@ -118,10 +116,16 @@ class MissionAssembler(private val id: Int) {
...
+        val homeCoordinates = nodes.first().coordinates3D
```
TimonSchreiber:
```suggestion
        val home = nodes.firstOrNull() as? MissionNode.Home
            ?: return ItemAssemblyResult.UnsupportedCommand
```

app/src/main/java/org/WenuLink/mavlink/controllers/NavigationController.kt
TimonSchreiber:
```suggestion
                logger.e { "Unsupported mission command: ${itemMsg.command}" }
```

TimonSchreiber (changes requested):
I have two small details and one suggestion for you, otherwise this is a good improvement.

By the way, have you tested what happens if the home position set in the planned mission is not the same as the drones current position and there are no takeoff coordinates?
Does the drone fly up first and only when reaching the desired takeoff altitude fly horizontally to the home coordinates? Or is there a risk that the drone tries to reach the x- and y-coordinates first and only then fly upwards?

---

The two minor things:

First there is a trailing "f" at the end of a comment on line 122.

Also upgrading the log levels seems reasonable, but if the `UnsupportedFrame`-case gets logged at error level, so should the `UnsupportedCommand` case.

---

Suggestion:

The fallback to known values for the x- and y-coordinates is good but I would suggest to use a nullsafe access method of the first node. Currently this list is guaranteed to have at least one element, but that relies on a different class and if that ever changes we suddenly get an exception in the `assembleNode()` method.

Also checking for the value of `0.0` individually seems a little bit risky, flying a drone on the equator or the prime meridian is not unreasonable, at (0.0, 0.0) it is. So similar to the `MissionManager.uploadMission()` method, I would suggest to check both values together and only then fall back to the home coordinates.

Looking at the MAVLink documentation, I can see that `Int.MAX_VALUE` is also a possible value if the paramater is not defined, so checking both `0.0` and `Int.MAX_VALUE` might be a cheap adjustment to avoid future problems.

[MAVLink/MISSION_ITEM_INT](https://mavlink.io/en/messages/common.html#MISSION_ITEM_INT)

angel-ayala:
hi @TimonSchreiber , thanks for your quick reply.

first regards to your questions, I didn't considered any of those cases during the tests, but according as the takeoff is being handled by [DJIAction](https://developer.dji.com/api-reference/android-api/Components/Missions/DJIGoToAction.html#djigotoaction_initwithcoordinatewithaltitude_inline), I think it will try to adjust all three coordinates simultaneously I can check it and let you know. Similarly happens with [DJIWaypoint](https://developer.dji.com/api-reference/android-api/Components/Missions/DJIWaypoint.html#djiwaypoint_altitude_inline).
I can test it and confirm it anyways.

Regards to your suggestions and considerations, all are being included. Regards to the `0.0` related condition, maybe there is a miss conception of such values, maybe first must check the frame type to later check on parameters validity and decides how to process the mission item. I began to think that the mission item is transporting `0.0` as coordinates since is possibly addressing a local frame, hence the origin of the aircraft more than a new location were to go.

If the aircraft is in another location, maybe the mission item should contain global coordinates instead. I think that should be a better approach to handle this, and must be discussed in another PR or issue.

TimonSchreiber (approved):
Good work!

My impression was that these are global coordinates. If they are local I wonder what the local (0, 0) coordinate would be? Home? Current position? Either way, falling back to a default value woudn't even be necessary then, right?

angel-ayala:
Exactly, at least for the ardupilot's firmware the 0, 0 position in local coordinates is where the EKF origin was set, not necessarily should be the home or current position.
However, I think that we could consider (0,0) the home position instead, since we don't have any EKF to set the origin.

